PR #24, opened 2023-06-16
feat(transport): add DNS-over-TCP PacketListener

jyyi1:


transport/dnsovertcp/packet_listener.go
```diff
@@ -0,0 +1,349 @@
...
+		return 0, errClosed
+	}
+	if udpAddr, ok := addr.(*net.UDPAddr); !ok || udpAddr.Port != dnsServerPort {
+		return 0, errInvalidResolver
+	}
+	if len(p) < dnsUdpMinMsgLen {
+		return 0, errInvalidDNSMsg
+	}
+
+	conn, n, err := c.sendDNSRequestOverTCP(p, addr)
```
fortuna, line 190: This waits on a TCP connection and will cause head-of-line blocking. You won't be able to process the next UDP packet until this write returns, while you could be doing that in parallel.

Move the Dial+Write to the Go routine. You already have the `buf` in `listenDNSResponseOverTCP`. Copy `p` there before starting the go-routine, and use it for the Write to the proxy.

fortuna (changes requested):
I made some suggestions to improve the implementation, but given the change in performance characteristics, we need to understand what we are getting out of it to justify this change.

transport/dnsovertcp/packet_listener.go
```diff
@@ -0,0 +1,349 @@
...
+	}()
+
+	if err = conn.SetReadDeadline(c.recvDdl.Deadline()); err != nil {
+		return
+	}
+	if err = conn.SetWriteDeadline(*c.wrDdl.Load()); err != nil {
+		return
+	}
+
+	// DoT: message length prefix
```
fortuna, line 302: DoT usually refers to DNS-over-TLS:
```suggestion
	// DNS-over-TCP: message length prefix
```

transport/dnsovertcp/packet_listener.go
fortuna, line 1: Here is another critical architecture choice: this logic should probably be in a lwip2transport UDP handler.
We may want to define a type for that.

That allows us to [return an error on the Connect call](https://github.com/eycorsican/go-tun2socks/blob/301549c435ff1ca8fc353bee3763ae639c727b89/proxy/dnsfallback/udp.go#L29), which we can't with this approach. I believe that may have implications on what the lwip sends to the TUN device (ICMP error?)

Implementing the handler also removes the need to implement a PacketConn, which makes the logic more complicated.

fortuna (changes requested):

transport/dnsovertcp/packet_listener.go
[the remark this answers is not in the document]
jyyi1: Here are the two call sequences. So in this PL, we don't need to depend on DNS client's retry logic, we will do the DNS-over-TCP for them.

#### Old DNS Fallback PL (See #26)

Sequence:

* `(1)`: DNS request over UDP
* `(2)`: In `WriteTo`, send the request to `ReadFrom` through an internal channel
* `(3)`: In `ReadFrom`, set the "TC" (truncated) bit (see #26) and return
* `(4)`: DNS client retry it over TCP

The DNS Client needs to implement the retry-over-TCP logic when it sees the "TC" (truncated) bit set in the UDP response.

```
    DNS Client (e.g. browser)   -----(4)----->   shadowsocks.StreamDialer
             |        ^
             |        |
            (1)       +-----------------(3)------------------+
             |                                               |
             v                                               |
dnstruncate.PacketListener.WriteTo(...)   --(2)-->  dnstruncate.PL.ReadFrom(...)
```

#### This PL (DNS-over-TCP)

Sequence:

* `(1)`: DNS request over UDP
* `(2)`: In `WriteTo`, wrap the DNS request in a TCP packet (by using `shadowsocks.StreamDialer`)
* `(3)`: The `shadowsocks.StreamDialer` tells the remote shadowsocks server to send DNS request over TCP and get the response
* `(4)`: In `ReadFrom`, read the DNS response (over TCP) from the tcp connection
* `(5)`: Unpack the TCP packet and convert it to a UDP DNS response, and send the response back to the Client

The DNS Client doesn't even need to know anything behind the scene, from its perspective, the UDP DNS requests successfully responded with a UDP DNS response.

```
    DNS Client (e.g. browser)
             |        ^
             |        |
            (1)       +-----------------(5)------------------+
             |                                               |
             v                                               |
dnsovertcp.PacketListener.WriteTo(...)             dnstovertcp.PL.ReadFrom(...)
                                |                                   ^
                                |                                   |
                               (2)              +--------(4)--------+
                                |               |
                                v               |
                              shadowsocks.StreamDialer  <--(3)-->  DNS Resolver (e.g. 8.8.8.8)
```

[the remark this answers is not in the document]
fortuna: I'm not worried about the client, because the client here is the operating system, which knows how to handle DNS properly.

Implementing a DNS client is complicated, so we can leverage the fact that the OS already provides the implementation. For instance, your client doesn't consider [EDNS0](https://datatracker.ietf.org/doc/html/rfc6891), which allows the client to specify a larger buffer size.

I found some data on the truncation bit:
image
From [Fragmentation, truncation, and timeouts: are large DNS messages falling to bits?](https://www.sidnlabs.nl/downloads/4nEIOFHKbAStsWQhh2l4Lr/5e69fe630bc7290713eb9638c2229828/Fragmentation__truncation__and_timeouts_are_large_DNS_messages_falling_to_bits.pdf) ([blog post](https://labs.ripe.net/author/giovane_moura/fragmentation-truncation-and-timeouts-are-large-dns-messages-falling-to-bits/)).

Around 3-7% of DNS responses had the TC bit set. However, that considers the use of EDNS0. Without it, the retries would happen a lot more. It looks like 40% of more of the UDP responses were larger than 512 bytes: 
image

By implementing the DNS-over-TCP logic we are forcing two double TCP connections (2 x (client-proxy + proxy-resolver)), effectively 8 round trips instead of 4 (2 for UDP), for 40% of the resolutions, which is substantial.

Given the added complexity, and performance regression, I don't think us and others will use this logic.

PR #27, opened 2023-07-05
feat(network): introduce PacketProxy to handle UDP traffic

jyyi1:
This PR introduces a network stack neutral `network.PacketProxy` that can be used to **relay** (e.g. a UDP proxy server) or **process** (e.g. a local DNS server) UDP packets. It is more efficient than the `transport.PacketListener` when you try to implement a local UDP server that can process UDP requests and get the responses without going to the internet (see #26 ).


network/packet_listener_handler.go
```diff
@@ -0,0 +1,98 @@
...
+var _ PacketHandler = (*packetListenerHandlerAdapter)(nil)
+var _ PacketSession = (*packetListenerSessionAdapter)(nil)
+
+type packetListenerHandlerAdapter struct {
+	transport.PacketListener
+}
+
+type packetListenerSessionAdapter struct {
+	net.PacketConn
+	response     PacketResponseWriter
```
fortuna, line 37: This will increase the readability. `response` is ambiguous, so is `writer`.

```suggestion
	responseWriter     PacketResponseWriter
```

network/packet_listener_handler.go
```diff
@@ -0,0 +1,98 @@
...
+func NewPacketHandlerFromPacketListener(pl transport.PacketListener) (PacketHandler, error) {
+	if pl == nil {
+		return nil, errors.New("pl must not be nil")
+	}
+	return &packetListenerHandlerAdapter{pl}, nil
+}
+
+// NewSession implements [PacketHandler].NewSession function. It uses [transport.PacketListener].ListenPacket to create
+// a [net.PacketConn], and constructs a new [PacketSession] that is based on this [net.PacketConn].
+func (plh *packetListenerHandlerAdapter) NewSession(laddr net.Addr, w PacketResponseWriter) (PacketSession, error) {
```
fortuna, line 53: `w` is ambiguous.

```suggestion
func (plh *packetListenerHandlerAdapter) NewSession(laddr net.Addr, responseWriter PacketResponseWriter) (PacketSession, error) {
```

network/packet_listener_handler.go
```diff
@@ -0,0 +1,98 @@
...
+func (s *packetListenerSessionAdapter) Close() error {
+	err := s.PacketConn.Close()
+	<-s.listenerDone
+	return err
+}
+
+// listenPackets reads packets from the [net.PacketConn] created by the [transport.PacketListener] until ReadFrom
+// returns [io.EOF]. s.Close() will make ReadFrom returns [io.EOF] as well.
+func (s *packetListenerSessionAdapter) listenPackets() {
+	defer close(s.listenerDone)
```
fortuna, line 87: Call `responseWriter.Close()` instead.
That will let handlers dictate the session duration and allow for reuse of that logic.

The closing of the response writer should trigger the closing of the request writer by the caller (lwip handler).

network/packet_listener_handler.go
```diff
@@ -0,0 +1,98 @@
...
+// Compilation guard against interface implementation
+var _ PacketHandler = (*packetListenerHandlerAdapter)(nil)
+var _ PacketSession = (*packetListenerSessionAdapter)(nil)
+
+type packetListenerHandlerAdapter struct {
+	transport.PacketListener
+}
+
+type packetListenerSessionAdapter struct {
+	net.PacketConn
```
fortuna, line 36: Using these embedded type makes readability harder.
```suggestion
	pc net.PacketConn
```

network/packet_listener_handler.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+const packetMaxSize = 2048
+
+// Compilation guard against interface implementation
+var _ PacketHandler = (*packetListenerHandlerAdapter)(nil)
+var _ PacketSession = (*packetListenerSessionAdapter)(nil)
+
+type packetListenerHandlerAdapter struct {
+	transport.PacketListener
```
fortuna, line 32: For readability:
```suggestion
	listener transport.PacketListener
```

network/lwip2transport/udp.go
```diff
@@ -15,103 +15,95 @@
...
+	// Lock the entire function to prevent from creating multiple sessions of one laddr
+	h.mu.Lock()
+	defer h.mu.Unlock()
+
+	if _, ok := h.sessions[laddr]; ok {
+		return fmt.Errorf("duplicated connection %v", laddr)
+	}
+
+	// Try to create a new session
+	respw := &udpConnResponseWriter{tunConn}
```
fortuna, line 64: We need a response writer that will remove the association on `Close()`.

network/lwip2transport/udp.go
```diff
@@ -15,103 +15,95 @@
...
+	session, err := h.handler.NewSession(tunConn.LocalAddr(), respw)
```
fortuna, line 65:
```suggestion
	requestWriter, err := h.handler.NewSession(tunConn.LocalAddr(), respw)
```

network/lwip2transport/udp.go
fortuna, line 64: For clarity:
```suggestion
	responseWriter := &udpConnResponseWriter{tunConn}
```

network/lwip2transport/udp.go
```diff
@@ -15,103 +15,95 @@
...
-		if err != nil {
-			return
-		}
-		_, err = tunConn.WriteFrom(buf[:n], sourceUDPAddr)
-		if err != nil {
-			return
-		}
-	}
+	// Clean-up the session after timeout
+	time.AfterFunc(h.timeout, func() {
```
fortuna, line 73: This is no longer an idle timeout. You are setting a fixed time out and never updating it. It will remove sessions on the fixed timeout regardless of use. That's not what we want.

The original logic was updating the timeout on every packet. It was measuring idle timeout. We need to restore that behavior.

I believe a good approach would be to put that logic in a handler though. On request (or response), you update the timeout. When the timeout expires, we close the ResponseWriter. That should tell the lwip handler to remove the association.

With that, people can provide their own session expiration logic. For example, in the server we expire DNS requests faster than other connections, since we have many of them. We could possibly reuse the same logic here.

fortuna (changes requested):

network/packet_handler.go
```diff
@@ -0,0 +1,64 @@
...
+	// PacketResponseWriter.Write to send UDP response packets.
+	NewSession(laddr net.Addr, w PacketResponseWriter) (PacketSession, error)
+}
+
+// PacketSession represents a UDP socket session. It can be used to send and receive UDP packets. To send a UDP packet,
+// call the WriteRequest function. To receive a UDP packet, the PacketSession will use the PacketResponseWriter object
+// passed to the PacketHandler.NewSession function.
+//
+// Multiple goroutines may invoke methods on a PacketSession simultaneously.
+type PacketSession interface {
```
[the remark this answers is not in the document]
jyyi1: Renamed to `PacketRequestSender`

network/lwip2transport/udp.go
jyyi1: Implemented timeout in `packetListenerProxyAdapter`

network/lwip2transport/udp.go
```diff
@@ -15,103 +15,110 @@
...
-	if proxyConn, ok := h.conns[laddr]; ok {
-		proxyConn.Close()
-		delete(h.conns, laddr)
+// Write relays packets from the proxy to the lwIP TUN device. It's called by transport.PacketSession.
+func (r *udpConnResponseWriter) WriteFrom(p []byte, source net.Addr) (int, error) {
+	srcAddr, err := net.ResolveUDPAddr("udp", source.String())
+	if err != nil {
+		return 0, err
 	}
+	return r.WriteFrom(p, srcAddr)
```
jyyi1, line 117:
```suggestion
	return r.conn.WriteFrom(p, srcAddr)
```

network/packet_listener_proxy.go
```diff
@@ -0,0 +1,97 @@
...
+	conn, err := proxy.listener.ListenPacket(context.Background())
+	if err != nil {
+		return nil, err
+	}
+	reqSender := &packetListenerRequestSender{conn, proxy.timeout}
+
+	// Read UDP responses asynchronously
+	go func() {
+		defer respWriter.Close()
+		buf := make([]byte, 0, packetMaxSize)
```
jyyi1, line 70:
```suggestion
		buf := make([]byte, packetMaxSize)
```

network/packet_listener_proxy.go
```diff
@@ -0,0 +1,97 @@
...
+// NewPacketProxyFromPacketListener creates a new [PacketProxy] that uses the existing [transport.PacketListener].
+// You can use this function if you already have an implementation of [transport.PacketListener] and would like to
+// inject it into one of the network stacks (for example, network/lwip2transport) as UDP traffic handlers.
+func NewPacketProxyFromPacketListener(pl transport.PacketListener) (PacketProxy, error) {
+	if pl == nil {
+		return nil, errors.New("pl must not be nil")
+	}
+	return &packetListenerProxyAdapter{
+		listener: pl,
+		timeout:  30 * time.Second,
```
fortuna, line 51: Not for this PR, but we may want to consider a strategy similar to the server.

We use a default of 5 minutes:
https://github.com/Jigsaw-Code/outline-ss-server/blob/989add59c0adb7d44013298f64f7eabdba45371e/cmd/outline-ss-server/main.go#L49

However, for DNS we use 17 seconds:
https://github.com/Jigsaw-Code/outline-ss-server/blob/989add59c0adb7d44013298f64f7eabdba45371e/service/udp.go#L275

Please read "Mapping Refresh" on RFC 4787: https://datatracker.ietf.org/doc/html/rfc4787#section-4.3

network/packet_listener_proxy.go
```diff
@@ -0,0 +1,97 @@
...
+		return nil, err
+	}
+	reqSender := &packetListenerRequestSender{conn, proxy.timeout}
+
+	// Read UDP responses asynchronously
+	go func() {
+		defer respWriter.Close()
+		buf := make([]byte, 0, packetMaxSize)
+		for {
+			conn.SetReadDeadline(time.Now().Add(proxy.timeout))
```
fortuna, line 72: I don't think we need to update the deadline on incoming packets. See https://datatracker.ietf.org/doc/html/rfc4787#section-4.3.

"NAT Inbound refresh behavior" actually allows an external party to keep the association forever, which can be an issue.

network/lwip2transport/udp.go
```diff
@@ -15,103 +15,110 @@
...
-	laddr := tunConn.LocalAddr().String()
-	tunConn.Close()
-	h.Lock()
-	defer h.Unlock()
-	if proxyConn, ok := h.conns[laddr]; ok {
-		proxyConn.Close()
-		delete(h.conns, laddr)
+// Write relays packets from the proxy to the lwIP TUN device. It's called by transport.PacketSession.
+func (r *udpConnResponseWriter) WriteFrom(p []byte, source net.Addr) (int, error) {
+	srcAddr, err := net.ResolveUDPAddr("udp", source.String())
```
fortuna, line 113: Why is this resolution needed? 
`r.WriteFrom` takes a `net.Addr`
Please add a comment.

network/lwip2transport/udp.go
```diff
@@ -15,103 +15,110 @@
...
+	if err != nil {
+		return 0, err
 	}
+	return r.WriteFrom(p, srcAddr)
+}
+
+func (r *udpConnResponseWriter) Close() error {
+	err := r.conn.Close()
```
fortuna, line 121: Perhaps move this to `closeSession`?
Or pass the local address `closeSession` to make it clear it doesn't use the conn

network/lwip2transport/udp.go
```diff
@@ -15,103 +15,110 @@
...
+	laddr := tunConn.LocalAddr().String()
+
+	h.mu.Lock()
+	reqSender, ok := h.senders[laddr]
+	h.mu.Unlock()
+
+	if !ok {
+		return fmt.Errorf("no session found for local address %v", laddr)
+	}
+	_, err := reqSender.WriteTo(data, destAddr)
```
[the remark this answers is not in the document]
fortuna: I'm still trying to figure out all the mux situation, and closing sequence. Perhaps we need a diagram of sorts. But I'll send the review for now.

fortuna (changes requested):
APIs look good. The comments are now on the implementation. By the way, you may want to read up on NATs. This is a great article: https://tailscale.com/blog/how-nat-traversal-works/
I also point to sections of an RFC in the comments.

network/packet_listener_proxy.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"context"
+	"errors"
+	"net"
+	"time"
+
+	"github.com/Jigsaw-Code/outline-internal-sdk/transport"
+)
+
+const packetMaxSize = 2048
```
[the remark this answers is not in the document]
jyyi1: Yes were using that by allocating a buffer from the [lwIP's buffer pool](https://github.com/Jigsaw-Code/outline-internal-sdk/blob/main/network/lwip2transport/udp.go#L72C31-L72C31), and [`lwip.BufSize` is 2048](https://github.com/eycorsican/go-tun2socks/blob/301549c435ff1ca8fc353bee3763ae639c727b89/core/buffer_pool.go#L9).

Though the RFC defined maximum UDP packet size is 64K. And QUIC's MTU is ~1600. So I think 2048/4096 might be good for most of the daily uses.

network/packet_listener_proxy.go
```diff
@@ -0,0 +1,97 @@
...
+
+	// Read UDP responses asynchronously
+	go func() {
+		defer respWriter.Close()
+		buf := make([]byte, 0, packetMaxSize)
+		for {
+			conn.SetReadDeadline(time.Now().Add(proxy.timeout))
+			n, addr, err := conn.ReadFrom(buf[:])
+			if err != nil {
+				return
```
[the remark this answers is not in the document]
jyyi1: Yes this was what we were doing: https://github.com/Jigsaw-Code/outline-internal-sdk/blob/main/network/lwip2transport/udp.go#L80-L82

Also I agree that we should ignore recoverable errors, but it's somehow hard to tell which one is recoverable here. I would prefer to be more specific about the errors we will ignore, for example `if !errors.Is(err, net.ErrShortBuffer) { return }` instead of `if errors.Is(err, io.EOF) || errors.Is(err, os.ExceededTimeout) { return }`.

network/packet_listener_proxy.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}()
+
+	return reqSender, nil
+}
+
+// WriteTo implements [PacketRequestSender].WriteTo function. It simply forwards the packet to the underlying
+// [net.PacketConn].WriteTo function.
+func (s *packetListenerRequestSender) WriteTo(p []byte, destination net.Addr) (int, error) {
+	s.conn.SetWriteDeadline(time.Now().Add(s.timeout))
```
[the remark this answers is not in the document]
jyyi1: Or we can simply use `s.conn.SetDeadline()` which will set both the read and write deadline.

network/lwip2transport/udp.go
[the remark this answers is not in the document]
jyyi1: I feel it's OK, we should allow `Close` and `WriteTo` to be called simultaneously. Because for example, from `net.PacketConn`:

> Multiple goroutines may invoke methods on a PacketConn simultaneously.

The same requirement should be applied to `PacketRequestSender` as well. So the implementor needs to take care of this race condition. And since our `PacketListenerProxy` is already using a `PacketConn`, so this is taken care of automatically.

network/lwip2transport/udp.go
```diff
@@ -15,103 +15,110 @@
...
+// ReceiveTo relays packets from the lwIP TUN device to the proxy. It's called by lwIP.
+func (h *udpHandler) ReceiveTo(tunConn lwip.UDPConn, data []byte, destAddr *net.UDPAddr) error {
+	laddr := tunConn.LocalAddr().String()
+
+	h.mu.Lock()
+	reqSender, ok := h.senders[laddr]
+	h.mu.Unlock()
+
+	if !ok {
+		return fmt.Errorf("no session found for local address %v", laddr)
```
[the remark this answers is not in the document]
jyyi1: Yup, then we don't even need to have any actual code in `Connect()`.

network/lwip2transport/udp.go
```diff
@@ -15,103 +15,110 @@
...
+func (h *udpHandler) Connect(tunConn lwip.UDPConn, _ *net.UDPAddr) error {
+	return h.newSession(tunConn)
+}
+
+// ReceiveTo relays packets from the lwIP TUN device to the proxy. It's called by lwIP.
+func (h *udpHandler) ReceiveTo(tunConn lwip.UDPConn, data []byte, destAddr *net.UDPAddr) error {
+	laddr := tunConn.LocalAddr().String()
+
+	h.mu.Lock()
+	reqSender, ok := h.senders[laddr]
```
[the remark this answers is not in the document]
jyyi1: It might help if we move the `NewSession` logic into `ReceiveTo`.

network/lwip2transport/udp.go
jyyi1: It should be `r.conn.WriteFrom`, which takes a `*net.UDPAddr`. will fix the code in next commit.

network/lwip2transport/udp.go
jyyi1: Prefer the first one.

network/packet_listener_proxy.go
jyyi1: This just makes sure `ReadFrom` will return after this timeout. If I understand it correctly, it should not affect anything about the NAT mapping?

network/lwip2transport/udp.go
jyyi1: Try `transport.MakeNetAddr`.

network/doc.go
```diff
@@ -14,9 +14,10 @@
 
 /*
 The network package defines interfaces and provides utilities for network layer (OSI layer 3) functionalities. For
-example, the IPDevice interface can be used to read and write IP packets from a physical or virtual network device.
+example, the [IPDevice] interface can be used to read and write IP packets from a physical or virtual network device.
 
 In addition, user-space network stack implementations are also included in the sub-packages (such as
-network/lwip2transport) that can translate raw IP packets into TCP/UDP connections.
+network/lwip2transport) that can translate raw IP packets into TCP/UDP connections. You can implement a [PacketProxy]
```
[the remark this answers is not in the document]
jyyi1: sounds good

network/lwip2transport/udp.go
jyyi1: `transport.MakeNetAddr` returns a `net.Addr`, but we need a `*net.UDPAddr` parsed from `net.Addr`. Then we still need to convert it to `*net.UDPAddr` by using type assertion, it doesn't simplify the code here, so I will just keep it as it is now.

fortuna (changes requested):
Looking good. I have mostly minor comments, the biggest question is whether we actually need the deadline timer.

internal/ddltimer/ddltimer.go
```diff
@@ -0,0 +1,125 @@
...
+// SetDeadline changes the timer to expire after deadline t. When the timer expires, the Timeout() channel will be
+// unblocked. A zero value means the timer will not time out.
+//
+// SetDeadline is like [time.Timer]'s Reset() function, but it doesn't have any restrictions.
+func (d *DeadlineTimer) SetDeadline(t time.Time) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	// Stop the timer, and if d.t has already invoked the callback of AfterFunc, create a new channel.
+	if d.t != nil && !d.t.Stop() {
```
fortuna, line 75: I'll point out that the timer we need is for a one-time expiration (unlike net.Conn's). We don't need to handle resets after the initial expiration, which simplifies things.

jyyi1: Yes, this is valid argument. I can try `AfterFunc` and `Reset` to see whether it works. But `NewTimer` and `Reset` is very hard to be used in multi-thread environment.

network/packet_listener_proxy.go
```diff
@@ -0,0 +1,115 @@
...
+	go func() {
+		reqSender.timer.SetDeadline(time.Now().Add(proxy.timeout))
+		<-reqSender.timer.Timeout()
+		reqSender.Close()
+	}()
+
+	// Relay incoming UDP responses from the proxy asynchronously until EOF or error
+	go func() {
+		defer respWriter.Close()
+		buf := make([]byte, packetMaxSize)
```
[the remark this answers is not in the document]
jyyi1: Checked go compiler, it will escape to heap, so updated to use `slicepool`.

internal/ddltimer/ddltimer.go
jyyi1: updated to use `time.AfterFunc`.

network/packet_listener_proxy.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	reqSender := &packetListenerRequestSender{
+		adapter:   proxy,
+		proxyConn: proxyConn,
+		timer:     ddltimer.New(),
+	}
+
+	// Terminate the session after timeout (deadline can be updated by WriteTo)
+	go func() {
+		reqSender.timer.SetDeadline(time.Now().Add(proxy.timeout))
```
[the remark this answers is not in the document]
jyyi1: I'll leave it as it is for now, since `Close()` is idempotent, it doesn't hurt to call it multiple times.

network/lwip2transport/device.go
```diff
@@ -47,29 +46,29 @@ type lwIPDevice struct {
...
+// But keep in mind that only one goroutine can call Write at a time; and only one goroutine can use either Read or
+// WriteTo at a time.
 //
 // [lwIP library]: https://savannah.nongnu.org/projects/lwip/
-func ConfigureDevice(sd transport.StreamDialer, pl transport.PacketListener) (network.IPDevice, error) {
-	if sd == nil || pl == nil {
-		return nil, errors.New("both sd and pl are required")
+func ConfigureDevice(sd transport.StreamDialer, pp network.PacketProxy) (network.IPDevice, error) {
+	if sd == nil || pp == nil {
+		return nil, errors.New("both sd and pktProxy are required")
```
fortuna, line 71:
```suggestion
		return nil, errors.New("both sd and pp are required")
```

network/packet_listener_proxy.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+	}()
+
+	return reqSender, nil
+}
+
+// WriteTo implements [PacketRequestSender].WriteTo function. It simply forwards the packet to the underlying
+// [net.PacketConn].WriteTo function.
+func (s *packetListenerRequestSender) WriteTo(p []byte, destination net.Addr) (int, error) {
+	s.writeIdleTimer.Reset(s.writeIdleTimeout)
```
fortuna, line 111: There's a race condition: if the timer expires, and you call Reset, it will be scheduled again. So we may end up with two calls to Close.

I think you may need to do the trick from the documentation:
https://pkg.go.dev/time#Timer.Reset
```go
if !t.Stop() {
	<-t.C
}
t.Reset(d)
```

fortuna: Also, it is safe to call Reset concurrently?

network/packet_listener_proxy.go
```diff
@@ -0,0 +1,120 @@
...
+	return s.proxyConn.WriteTo(p, destination)
+}
+
+// Close implements [PacketRequestSender].Close function. It closes the underlying [net.PacketConn]. This will also
+// terminate the goroutine created in NewSession because s.conn.ReadFrom will return [io.EOF].
+func (s *packetListenerRequestSender) Close() error {
+	s.writeIdleTimer.Stop()
```
fortuna, line 118: Do you need to drain the channel, as described in https://pkg.go.dev/time#Timer.Stop?

fortuna (commented):
Looks good. I have a couple of questions on the timer

network/packet_listener_proxy.go
jyyi1: The GoDoc doesn't specify anything about thread-safety of `Timer`, I'll just assume it is NOT thread safe. So I added a `Mutex` to protect timer function calls. I also updated the `Close()` call to check the private `closed` first. Since `Close()` is idempotent, calling it multiple times is OK, it's OK to have the timer's race condition here.

network/packet_listener_proxy.go
jyyi1: I don't think so, [`AfterFunc` doesn't have the `C` initialized](https://cs.opensource.google/go/go/+/refs/tags/go1.20.6:src/time/sleep.go;l=164-170), it will cause deadlock if you do so (check this in the playground: https://go.dev/play/p/RK_zbuhNUdQ)

network/packet_listener_proxy.go
```diff
@@ -108,13 +114,26 @@ func (proxy *packetListenerProxyAdapter) NewSession(respWriter PacketResponseRec
 // WriteTo implements [PacketRequestSender].WriteTo function. It simply forwards the packet to the underlying
 // [net.PacketConn].WriteTo function.
 func (s *packetListenerRequestSender) WriteTo(p []byte, destination net.Addr) (int, error) {
+	s.mu.Lock()
```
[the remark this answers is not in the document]
jyyi1: We can limit the locking scope to exclude `proxyConn.WriteTo`. `proxyConn` is thread-safe.

network/errors.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+// Portable analogs of some common errors.
+//
+// Errors returned from this package and all sub-packages may be tested against these errors with [errors.Is].
+
+// ErrClosed is the error returned by an I/O call on a network device that has already been closed, or that is closed
+// by another goroutine before the I/O is completed. This may be wrapped in another error, and should normally be
+// tested using errors.Is(err, network.ErrClosed).
+var ErrClosed = errors.Join(os.ErrClosed, net.ErrClosed, syscall.ESHUTDOWN)
```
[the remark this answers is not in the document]
jyyi1: My intention is that people can either test against our error `errors.Is(err, network.ErrClosed)` or test against one of the commonly used errors in go: `errors.Is(err, net.ErrClosed)` or `errors.Is(err, os.ErrClosed)`.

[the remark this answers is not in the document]
jyyi1: updated

network/packet_listener_proxy.go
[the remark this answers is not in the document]
jyyi1: updated

PR #54, opened 2023-08-23
connectivity test app (without android)

daniellacosse:
README: https://github.com/Jigsaw-Code/outline-sdk/tree/outline-connectivity-app/x/outline-connectivity-app

### DIFF
--- a/x/examples/outline-connectivity-app/README.md
+++ b/x/examples/outline-connectivity-app/README.md
@@ -119,6 +119,7 @@ If at any point you run into issues during development, try `yarn reset`.
 1. **\[P1\]** android (in progress)
 1. **\[P1\]** read browser language on load, centralize language list, and only localize once
 1. **\[P1\]** documentation on how to generate mobile app build credentials
+1. **\[P1\]** add individual test result errors to the test result output UI
 1. **\[P2\]** use x/config to parse the access key and showcase the different transports (see: https://github.com/Jigsaw-Code/outline-sdk/blob/main/x/examples/outline-connectivity/main.go)
 1. **\[P2\]** generalize request handler via generics/reflection
 1. **\[P2\]** Create a logo for the app


x/outline-connectivity-app/.gitignore
```diff
@@ -0,0 +1,7 @@
+output
+node_modules
+.yarn/cache
+.yarn/sdks
+.yarn/unplugged
+.yarn/install-state.gz
```
fortuna, line 6: This seems to be [Yarn's recommendation](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored):

```suggestion
.pnp.*
.yarn/*
!.yarn/patches
!.yarn/plugins
!.yarn/releases
!.yarn/sdks
!.yarn/versions
```

x/outline-connectivity-app/.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "typescript.tsdk": ".yarn/sdks/typescript/lib",
+  "typescript.enablePromptUseWorkspaceTsdk": true,
+  "search.exclude": {
```
fortuna, line 4: I have the impression vscode excludes what's in .gitignore, so we may not need this.

x/outline-connectivity-app/go.mod
```diff
@@ -0,0 +1,40 @@
+module outline_sdk_connectivity_test
```
fortuna, line 1: This should match the directory name

x/outline-connectivity-app/go.mod
```diff
@@ -0,0 +1,40 @@
...
+
+go 1.20
+
+require (
+	github.com/Jigsaw-Code/outline-sdk v0.0.2
```
fortuna, line 6: This is the latest:
```suggestion
	github.com/Jigsaw-Code/outline-sdk v0.0.4
```

x/outline-connectivity-app/go.mod
```diff
@@ -0,0 +1,40 @@
...
+	github.com/Jigsaw-Code/outline-sdk/x v0.0.0-20230807220427-893de7fdc6b8
```
fortuna, line 7: Update it to the latest main. I think you can just replace the version with `main` and call `go mod tidy`

x/outline-connectivity-app/package.json
```diff
@@ -0,0 +1,34 @@
...
+  },
+  "engines": {
+    "node": ">=18.0.0"
+  },
+  "name": "outline_sdk_connectivity_test",
+  "packageManager": "yarn@3.6.1",
+  "private": true,
+  "scripts": {
+    "app_desktop": "yarn workspace app_desktop",
+    "app_mobile": "yarn workspace app_mobile",
```
fortuna, line 19: What do these commands do? Are they shortcuts to run commands in those workspaces?

x/outline-connectivity-app/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "build": "yarn workspaces foreach -ptiv run build",
+    "clean": "yarn workspaces foreach -piv run clean && rm -rf node_modules .yarn/cache .yarn/sdks .yarn/unplugged .yarn/install-state.",
+    "reset": "yarn clean && yarn && yarn sdks vscode",
```
fortuna, line 22: What does `yarn sdks vscode` do?
Is it the same as `yarn dlx @yarnpkg/sdks vscode` (from https://yarnpkg.com/getting-started/editor-sdks#usage)?

x/outline-connectivity-app/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "forceConsistentCasingInFileNames": true,
+    "isolatedModules": true,
+    "moduleResolution": "Node",
+    "noFallthroughCasesInSwitch": true,
+    "noImplicitReturns": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "skipLibCheck": true,
+    "strict": true,
+    "target": "ES5",
```
fortuna, line 20: Do we need to target ES5?

x/outline-connectivity-app/tsconfig.json
```diff
@@ -0,0 +1,28 @@
+{
```
fortuna, line 1: Perhaps we should use Google's standard tsconfig: https://github.com/google/gts/blob/main/tsconfig-google.json

It's fine if we tweak it. Maybe do an import as in https://github.com/google/gts/blob/main/tsconfig.json

x/outline-connectivity-app/tsconfig.node.json
fortuna, line 1: What is this file for?

x/outline-connectivity-app/tsconfig.node.json
```diff
@@ -0,0 +1,12 @@
+{
```
fortuna, line 1: Perhaps it can also import the Google tsconfig?

fortuna (changes requested):
Initial review. Root directory only.

x/outline-connectivity-app/shared_backend/index.ts
```diff
@@ -0,0 +1,25 @@
...
+export function from(backend: Invokable) {
+  return {
+    connectivityTest: invocationFactory<ConnectivityTestInput, ConnectivityTestResult[]>(backend, "ConnectivityTest")
+  }
+}
+
+/* INFRASTRUCTURE (to be moved) */
+
+export interface Invokable {
+  Invoke(parameters: { method: string; input: string; }): Promise<{ result: string; errors: string[] }>
```
fortuna, line 12: The output error should be a single error, not a list. It's hard to reason about a list. It's definitely not a common practice.

The error would need to be a struct, with a message and an optional cause that we can unwrap if needed.

That will nicely map to the JS Error.

x/outline-connectivity-app/shared_backend/index.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+function invocationFactory<T, K>({ Invoke }: Invokable, methodName: string) {
+  return async (input: T): Promise<K> => {
+    const { result, errors } = await Invoke({ method: methodName, input: JSON.stringify(input) });
```
fortuna, line 17: If we inline the parameters object, the call will be more readable:
```suggestion
    const { result, errors } = await Invoke(methodName, JSON.stringify(input));
```

x/outline-connectivity-app/shared_backend/main.go
```diff
@@ -0,0 +1,292 @@
...
+/* INFRASTRUCTURE (to be generalized via reflection/generics and moved) */
+type CallInputMessage struct {
+	Method string `json:"method"`
+	Input  string `json:"input"`
+}
+
+type CallOutputMessage struct {
+	Result string   `json:"result"`
+	Errors []string `json:"errors"`
+}
```
fortuna, line 246: I found these names a bit counter-intuitive.

Perhaps name them `CallRequest` and `CallResponse` to be more in line with the typical RPC names.

x/outline-connectivity-app/shared_backend/main.go
```diff
@@ -0,0 +1,292 @@
...
+
+func SendRawCall(rawInputMessage []byte) []byte {
+	var inputMessage CallInputMessage
```
fortuna, line 249:
```suggestion
func SendRawCall(rawRequest []byte) []byte {
	var request CallRequest
```

x/outline-connectivity-app/shared_backend/main.go
```diff
@@ -0,0 +1,292 @@
...
+
+	parseInputError := json.Unmarshal(rawInputMessage, &inputMessage)
+
+	outputMessage := CallOutputMessage{Result: "", Errors: []string{}}
```
fortuna, line 253: No need to initialize the fields with zero values:
```suggestion
	var request CallRequest
```

x/outline-connectivity-app/shared_backend/main.go
```diff
@@ -0,0 +1,292 @@
...
+
+	if parseInputError != nil {
+		outputMessage.Errors = append(outputMessage.Errors, "SendRawCall: error parsing raw input string")
+	}
+
+	if inputMessage.Method != "ConnectivityTest" {
```
fortuna, line 259: Can we do the other way around?

I'd like to see what code is specific to `ConnectivityTest`, so I know how to add a new method.

x/outline-connectivity-app/shared_backend/main.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+		rawBytes[i] = byte(r)
+	}
+	return rawBytes, nil
+}
+
+/* INFRASTRUCTURE (to be generalized via reflection/generics and moved) */
+type CallInputMessage struct {
+	Method string `json:"method"`
+	Input  string `json:"input"`
```
fortuna, line 240: Rename `Input` to the language we use for functions, like "parameters" or "arguments". Could be "params" or "args".

x/outline-connectivity-app/app_desktop/app.ts
```diff
@@ -0,0 +1,20 @@
...
+import * as SharedFrontend from "shared_frontend";
+import { LitElement, html } from "lit";
+import { customElement } from "lit/decorators.js";
+
+SharedFrontend.registerAllElements();
+
+// main
+@customElement("app-main")
+export class AppMain extends LitElement {
+  backend = SharedBackend.from(DesktopBackend);
```
fortuna, line 15: Can we do this?
```suggestion
  backend = new DesktopBackend();
```

x/outline-connectivity-app/shared_backend/index.ts
```diff
@@ -0,0 +1,25 @@
+import { ConnectivityTestInput, ConnectivityTestResult } from "./generated/types"
+
+export function from(backend: Invokable) {
+  return {
+    connectivityTest: invocationFactory<ConnectivityTestInput, ConnectivityTestResult[]>(backend, "ConnectivityTest")
```
fortuna, line 5: It seems like someone would have to modify this code in order to add a new method.

It's not clear what are all the places where someone would need to touch to make a method a available, but I was hoping it would be a single place.

We are also introducing a significant amount of complexity by coupling the IPC logic with the specific service we are providing. It makes the logic harder to understand, extend and distracts from highlighting the SDK.

For an IPC system, I was hoping to see first a primitive just to pass messages around, like we see in Flutter, Electron and Cordova.

We should be able to write Typescript frontend code that reads like this:
```ts
  channel = new MethodChannel('my.namespace');  
  ...
  result, error = await channel.invoke('testConnectivity', {config: '...'});
```

That's a building block that is decoupled from the specific backend service we are providing and would make it easier to understand the IPC system.

The server side is trivial, as it's just dispatching based on the method name. To add a new method, just update the dispatcher.

Once we have that, we can then build upon it and add type safety. I can see a dispatcher that dispatches to different methods using reflection given an interface, and code to make the invoke calls can be generated from the same interface. That's pretty much what Google's gRPC does. There are established patterns for this.

Can you refactor the code so there's a clear IPC mechanism that is independent of the backend service? It's fine if we end up needing to drop type safety for now.

/cc @jyyi1 since we discussed this at length before and has thoughts too.

fortuna (changes requested):
Review for `app_desktop`, `shared_backend` and the IPC system.

x/examples/outline-connectivity-app/.pnp.cjs
jyyi1, line 1: Is this file auto generated?

x/examples/outline-connectivity-app/.pnp.loader.mjs
jyyi1, line 1: Same question, is this file auto generated? If so we may remove it.

x/examples/outline-connectivity-app/.vscode/settings.json
jyyi1, line 1: We need to decide whether we should check in vscode settings file to the repo.

x/examples/outline-connectivity-app/.yarn/plugins/@yarnpkg/plugin-typescript.cjs
jyyi1, line 1: Same question, is this auto generated?

x/examples/outline-connectivity-app/README.md
```diff
@@ -0,0 +1,125 @@
+# Outline Connectivity Test App
```
jyyi1, line 1:
```suggestion
# Outline Connectivity Test GUI App
```

x/examples/outline-connectivity-app/app_desktop/app.go
```diff
@@ -0,0 +1,41 @@
+package main
```
jyyi1, line 1: Please include the Jigsaw Copyright headers to the source code files.

x/examples/outline-connectivity-app/app_desktop/generated/wailsjs/go/main/App.d.ts
```diff
@@ -0,0 +1,5 @@
+// Cynhyrchwyd y ffeil hon yn awtomatig. PEIDIWCH Â MODIWL
+// This file is automatically generated. DO NOT EDIT
```
jyyi1, line 2: I think we should not check in auto-generated files.

x/examples/outline-connectivity-app/app_desktop/generated/wailsjs/go/main/App.js
```diff
@@ -0,0 +1,7 @@
+// @ts-check
+// Cynhyrchwyd y ffeil hon yn awtomatig. PEIDIWCH Â MODIWL
+// This file is automatically generated. DO NOT EDIT
```
jyyi1, line 3: Same as above, can we exclude auto-generated files?

x/examples/outline-connectivity-app/app_desktop/main.go
```diff
@@ -0,0 +1,38 @@
...
+
+	// Create application with options
+	err := wails.Run(&options.App{
+		Title:  "@outline_sdk_connectivity_test/app_desktop",
+		Width:  1024,
+		Height: 768,
+		AssetServer: &assetserver.Options{
+			Assets: assets,
+		},
+		BackgroundColour: &options.RGBA{R: 27, G: 38, B: 54, A: 1},
```
jyyi1, line 28: no need to resolve in this PR, can we set the background color in css?

x/examples/outline-connectivity-app/app_desktop/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "lit": "^2.7.6",
+    "shared_backend": "workspace:^",
+    "shared_frontend": "workspace:^"
+  },
+  "devDependencies": {
+    "concurrently": "^8.2.0",
+    "typescript": "^4.6.4",
+    "vite": "^3.0.7"
+  },
+  "name": "app_desktop",
```
jyyi1, line 12: Shall we name it as `outline-connectivity`?

x/examples/outline-connectivity-app/app_desktop/package.json.md5
jyyi1, line 1: Is this checksum file a standard nodejs feature?

x/examples/outline-connectivity-app/app_mobile/ios/.gitignore
jyyi1, line 1: *nit* I think we only need one `.gitignore` for one project.

x/examples/outline-connectivity-app/app_mobile/ios/App/App/Assets.xcassets/AppIcon.appiconset/AppIcon-512@2x.png
jyyi1, line 1: Probably not in this PR, maybe replace it with the Outline logo?

x/examples/outline-connectivity-app/app_mobile/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "App",
```
jyyi1, line 2: Shall we name it as "OutlineConnectivityApp"?

x/examples/outline-connectivity-app/app_mobile/package.json
```diff
@@ -0,0 +1,28 @@
...
+  },
+  "devDependencies": {
+    "@capacitor/cli": "latest",
+    "concurrently": "^8.2.0",
+    "typescript": "^5.1.6",
+    "vite": "^2.9.13"
+  },
+  "name": "app_mobile",
+  "scripts": {
+    "build": "concurrently 'yarn:build:*'",
```
jyyi1, line 18: Is `concurrently` a preinstalled command line? Or is it included in `node_modules`?

x/examples/outline-connectivity-app/go.mod
```diff
@@ -0,0 +1,40 @@
+module outline_sdk_connectivity_test
+
+go 1.20
+
+require (
+	github.com/Jigsaw-Code/outline-sdk v0.0.5-0.20230828133257-302b739dffcd
```
jyyi1, line 6: I think the latest version is `v0.0.4`, I prefer to use a released version of SDK.

x/examples/outline-connectivity-app/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "author": "Daniel LaCosse",
+  "dependencies": {
+    "lit": "^2.7.6"
+  },
+  "description": "A basic cross-platform demo of the Outline SDK",
```
jyyi1, line 6:
```suggestion
  "description": "A cross-platform connectivity test app using the Outline SDK",
```

x/examples/outline-connectivity-app/shared_backend/generated/types.ts
```diff
@@ -0,0 +1,45 @@
+// Code generated by tygo. DO NOT EDIT.
```
jyyi1, line 1: Same as above, can we exclude this `generated` folder from the repo?

x/examples/outline-connectivity-app/shared_backend/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "devDependencies": {
+    "concurrently": "^8.2.0",
+    "nodemon": "^3.0.1"
+  },
+  "module": "index.ts",
+  "name": "shared_backend",
```
jyyi1, line 7:
```suggestion
  "name": "OutlineConnectivityTestBackend",
```

x/examples/outline-connectivity-app/shared_frontend/assets/fonts/jigsaw_sans/bold.woff2
jyyi1, line 1: Can we use webfont here?

x/examples/outline-connectivity-app/shared_frontend/index.ts
```diff
@@ -0,0 +1,6 @@
+import "./theme.css";
+import { ConnectivityTestPage } from "./pages";
+
+export function registerAllElements() {
+  window.customElements.define("connectivity-test-page",ConnectivityTestPage);
```
jyyi1, line 5:
```suggestion
  window.customElements.define("connectivity-test-page", ConnectivityTestPage);
```

x/examples/outline-connectivity-app/shared_frontend/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "dependencies": {
+    "@lit/localize": "^0.11.4",
+    "lit": "^2.7.6",
+    "shared_backend": "workspace:^"
+  },
+  "devDependencies": {
+    "@lit/localize-tools": "^0.6.9"
+  },
+  "module": "index.ts",
+  "name": "shared_frontend",
```
jyyi1, line 11:
```suggestion
  "name": "OutlineConnectivityTestFrontend",
```

x/examples/outline-connectivity-app/shared_frontend/pages/connectivity_test/generated/messages/index.ts
```diff
@@ -0,0 +1,27 @@
+// Do not modify this file by hand!
```
jyyi1, line 1: Same thing, shall we exclude this `generated/` folder?

jyyi1 (commented):
Thanks for the contribution! It's very useful.

x/outline-connectivity-app/go.mod
jyyi1: `github.com/Jigsaw-Code/outline-sdk/x/outline-connectivity-app`

x/examples/outline-connectivity-app/shared_frontend/assets/fonts/jigsaw_sans/bold.woff2
jyyi1: Ah, i mean just a URL to the web font.

x/examples/outline-connectivity-app/.vscode/settings.json
jyyi1: I'm OK with that. Or we can mention it in README, it's just one or two lines of settings.

x/examples/outline-connectivity-app/shared_frontend/assets/fonts/jigsaw_sans/bold.woff2
fortuna: +1 to bundling every resource we need. Using external resources is a major security risk. Also, it breaks if the user is offline, or, in the case of Google fonts, it doesn't even work in China. We had that issue before and we needed to force a Polymer component to bundle the fonts before.

x/examples/outline-connectivity-app/README.md
```diff
@@ -0,0 +1,128 @@
...
+  A -.-> |gomobile| D
+  A --> G
+  B --> H
+  B --> K
+  C --> I
+  D --> J
+  I --> L
+  J --> L
+  L --> K
+  G --> H
```
jyyi1, line 50: Similar to the comment above, `<->` for two-way communication?

x/examples/outline-connectivity-app/app_mobile/index.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
```
jyyi1, line 1: Do we need to put copyright header to HTML files like [this](https://github.com/Jigsaw-Code/outline-client/blob/master/src/www/index_electron.html#L1)?

x/outline-connectivity-app/go.mod
jyyi1: has this been updated?

x/examples/outline-connectivity-app/shared_frontend/pages/connectivity_test/localize.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "https://raw.githubusercontent.com/lit/lit/main/packages/localize-tools/config.schema.json",
+  "sourceLocale": "en",
+  "targetLocales": ["es-419", "zh-Hans", "fa-IR"],
```
fortuna, line 4: I don't think you need to specify a country for Farsi.
```suggestion
  "targetLocales": ["es-419", "zh-Hans", "fa"],
```

x/examples/outline-connectivity-app/shared_frontend/pages/connectivity_test/index.ts
```diff
@@ -0,0 +1,802 @@
...
+
+      /* if ios */
+      .header {
+        padding-top: 3.7rem;
+      }
+    }
+  `;
+
+  render() {
+    return html`<main dir="${this.locale === "fa-IR" ? "rtl" : "ltr"}">
```
fortuna, line 582: This is not quite maintainable. Consider keeping a [centralized list of supported languages](https://github.com/Jigsaw-Code/outline-client/blob/0cae8d0fb1ac2706ba25dfd9ea93b657ca6db45b/src/www/ui_components/app-root.js#L442), or perhaps try using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getTextInfo

fortuna (commented):
Review for `shared_backend`

x/outline-connectivity-app/shared_backend/main.go
fortuna: Sorry, I wasn't clear. I meant I'd like to see something like this:
```
if inputMessage.Method == "ConnectivityTest" {
  return HandleConnectivityTest(request, response)
}
```

So it allows the addition of new methods.
Ideally in a way that we don't need to repeat the boilerplate for every method.
Perhaps you have something like:
 
```go
if inputMessage.Method == "ConnectivityTest" {
  return Call(HandleConnectivityTest, request, response)
}
``` 

You can use Generics like:
```go
fun Call[Request, Response](call func(Request, Response) error, request []bytes, response []bytes) error {
  // Unmarshall, call, Marshall
}
```

x/examples/outline-connectivity-app/shared_frontend/pages/connectivity_test/index.ts
fortuna: Yes

x/examples/outline-connectivity-app/shared_backend/main.go
```diff
@@ -0,0 +1,232 @@
...
+	Hostname  string
+	Port      int
+	CryptoKey *shadowsocks.EncryptionKey
+	Prefix    Prefix
+}
+
+type Prefix []byte
+
+func ConnectivityTest(request ConnectivityTestRequest) ([]ConnectivityTestResult, error) {
+	accessKeyParameters, err := parseAccessKey(request.AccessKey)
```
fortuna, line 80: This code is tied to Shadowsocks. I'd like to see this use the `x/config` package instead, so it can showcase the different transports. See https://github.com/Jigsaw-Code/outline-sdk/blob/main/x/examples/outline-connectivity/main.go.

But let's do that in a different PR.

fortuna (approved):
Code looks good. The IPC calls are a lot easier to understand, thanks for the changes.
There's a README comment that you can fix, and we should use the config library, but that can be in another PR.

fortuna (approved):

jyyi1 (approved):

PR #101, opened 2023-10-13
add padding to header only on ios

daniellacosse:
Screenshot 2023-10-16 at 2 06 28 PM


x/examples/outline-connectivity-app/shared_frontend/pages/connectivity_test/index.ts
```diff
@@ -742,22 +752,41 @@ export class ConnectivityTestPage extends LitElement {
     </main>`;
   }
 
+  renderPlatformStyles() {
+    if (!this.platform) {
+      return nothing;
+    }
+
+    // TODO: changes how switches look on android
+    switch (this.platform.operatingSystem) {
```
jyyi1, line 761: This looks a little bit inflexible. Is it possible to add a new class to the root element like `platform-ios`, and in the CSS we can have dedicated rule for this class?

jyyi1 (commented):
Code looks good, with a minor comment.

jyyi1 (approved):
LGTM, thanks!

PR #109, opened 2023-10-18
Several improvements to toolchain setup for GUI app 

amircybersec:
I suggest several changes to setup and build configuration for the Outline connectivity GUI app. cc: @daniellacosse

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,11 @@
 # MacOS files
 .DS_Store
 .idea
+
+.yarn/*
+!.yarn/cache
+!.yarn/patches
+!.yarn/plugins
+!.yarn/releases
+!.yarn/sdks
+!.yarn/versions
--- a/.gitignore
+++ b/.gitignore
@@ -4,10 +4,12 @@
 .DS_Store
 .idea
 
+# Not using Zero-Installs
 .yarn/*
-!.yarn/cache
-!.yarn/patches
-!.yarn/plugins
 !.yarn/releases
-!.yarn/sdks
-!.yarn/versions
+!.yarn/plugins
+.pnp.*
+
+#  using Zero Installs
+#.yarn/unplugged
+#.yarn/build-state.yml
--- a/x/examples/outline-connectivity-app/app_mobile/android/app/src/main/AndroidManifest.xml
+++ b/x/examples/outline-connectivity-app/app_mobile/android/app/src/main/AndroidManifest.xml
@@ -7,7 +7,9 @@
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
-        android:theme="@style/AppTheme">
+        android:theme="@style/AppTheme"
+        android:usesCleartextTraffic="true">
+        
 
         <activity
             android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|smallestScreenSize|screenLayout|uiMode"
@@ -38,4 +40,5 @@
     <!-- Permissions -->
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 </manifest>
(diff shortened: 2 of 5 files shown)


.gitignore
amircybersec, line 1: some more context on this [here](https://stackoverflow.com/questions/58550439/am-i-meant-to-commit-yarns-pnp-js-file)

x/examples/outline-connectivity-app/shared_backend/package.json
amircybersec, line 1: On Ubuntu `-o output/SharedBackend.aar` keeps failing with `No such file or directory` error. The problem is that in `bash`, directory must exist before attempting to create a file. This behavior might be different on Mac though. For example:

```
$ touch test/test.txt

touch: cannot touch 'test/test.txt': No such file or directory
```

To address this issue, I am first checking of directory `output` exists, and if not, it gets created during build: `[ ! -d output ] && mkdir output;`

x/examples/outline-connectivity-app/app_mobile/android/app/src/main/AndroidManifest.xml
amircybersec, line 1: Overall the app is going to make http requests sometimes for network debugging purposes. Since our use case is different, this probably does not pose a problem in production. In debugging, we need `android:usesCleartextTraffic="true"` to be able to communicate with vite server. `android:name="android.permission.ACCESS_NETWORK_STATE"` would be needed to get some more information about network. We can plan permissions more rigorously for release

amircybersec: @daniellacosse Just to clarify, you mean instead of directly editing `AndroidManifext.xml`, we have a higher level manifest file with parameters and settings that get pulled by each platform? In other words, have `AndroidManifext.xml` generated during build?

amircybersec: The standard way of specifying the permissions in Android is through `AndroidManifext.xml` at `android/app/src/main/AndroidManifest.xml`. I did some research on cross-platform frameworks (react-native, [capacitor](https://capacitorjs.com/docs/android/configuration), etc), and this is still the way to go with these frameworks. Since the connectivity app needs to make various types of network calls (including HTTP) for testing purposes, we need to allow that in the permissions since more recent versions of Android prevent that by default: https://koz.io/android-m-and-the-war-on-cleartext-traffic/

x/examples/outline-connectivity-app/shared_backend/package.json
amircybersec: fixed :)

x/examples/outline-connectivity-app/shared_backend/package.json
```diff
@@ -6,9 +6,9 @@
   "module": "index.ts",
   "name": "shared_backend",
   "scripts": {
-    "build": "concurrently 'yarn:build:*' --kill-others-on-fail",
-    "build:android": "gomobile bind -target android -androidapi 21 -o output/SharedBackend.aar github.com/Jigsaw-Code/outline-sdk/x/examples/outline-connectivity-app/shared_backend",
-    "build:ios": "gomobile bind -target ios,iossimulator -o output/SharedBackend.xcframework github.com/Jigsaw-Code/outline-sdk/x/examples/outline-connectivity-app/shared_backend",
+    "build": "concurrently 'yarn:build:*'",
+    "build:android": "[ ! -d output ] && mkdir -p output; gomobile bind -target android -androidapi 21 -o output/SharedBackend.aar github.com/Jigsaw-Code/outline-sdk/x/examples/outline-connectivity-app/shared_backend",
+    "build:ios": "[ ! -d output ] && mkdir -p output; gomobile bind -target ios,iossimulator -o output/SharedBackend.xcframework github.com/Jigsaw-Code/outline-sdk/x/examples/outline-connectivity-app/shared_backend",
```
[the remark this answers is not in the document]
amircybersec: got myself confused a bit. that's right. just made the change

PR #116, opened 2023-10-24
Reporter package

amircybersec:
I exasperated the send report functionality into a separate package to be imported when needed.

### DIFF
--- a/x/reporter/reporter.go
+++ b/x/reporter/reporter.go
@@ -17,6 +17,7 @@ package reporter
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -25,11 +26,12 @@ import (
 )
 
 var debugLog log.Logger = *log.New(io.Discard, "", 0)
+var httpClient = &http.Client{}
 
 func sendReport(record map[string]interface{}, collectorURL string) error {
 	jsonData, err := json.Marshal(record)
 	if err != nil {
-		log.Fatalf("Error encoding JSON: %s\n", err)
+		log.Printf("Error encoding JSON: %s\n", err)
 		return err
 	}
 
@@ -40,10 +42,9 @@ func sendReport(record map[string]interface{}, collectorURL string) error {
 	}
 
 	req.Header.Set("Content-Type", "application/json; charset=utf-8")
-	client := &http.Client{}
-	resp, err := client.Do(req)
+	resp, err := httpClient.Do(req)
 	if err != nil {
-		log.Fatalf("Error sending the HTTP request: %s\n", err)
+		log.Printf("Error sending the HTTP request: %s\n", err)
 		return err
 	}
 	defer resp.Body.Close()
@@ -58,6 +59,16 @@ func sendReport(record map[string]interface{}, collectorURL string) error {
 }
 
 func sendReportRandomly(record map[string]interface{}, collectorURL string, success bool, successSampleRate float64, failureSampleRate float64) error {
+
+	// Perform custom range validation for sampling rate
+	if successSampleRate < 0.0 || successSampleRate > 1.0 {
+		return errors.New("Error: successSampleRate must be between 0 and 1.")
+	}
+
+	if failureSampleRate < 0.0 || failureSampleRate > 1.0 {
+		return errors.New("Error: failureSampleRate must be between 0 and 1.")
+	}
+
 	var samplingRate float64
 	if success {
 		samplingRate = successSampleRate
@@ -70,7 +81,7 @@ func sendReportRandomly(record map[string]interface{}, collectorURL string, succ
 		// Run your function here
 		err := sendReport(record, collectorURL)
 		if err != nil {
-			log.Fatalf("HTTP request failed: %v", err)
+			log.Printf("HTTP request failed: %v", err)
 			return err
 		} else {
 			fmt.Println("Report sent")
--- a/x/reporter/reporter.go
+++ b/x/reporter/reporter.go
@@ -1,0 +1,83 @@
+// Copyright 2023 Jigsaw Operations LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package reporter
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"log"
+	"math/rand"
+	"net/http"
+)
+
+var debugLog log.Logger = *log.New(io.Discard, "", 0)
+
+func sendReport(record map[string]interface{}, collectorURL string) error {
+	jsonData, err := json.Marshal(record)
+	if err != nil {
+		log.Fatalf("Error encoding JSON: %s\n", err)
+		return err
+	}
+
+	req, err := http.NewRequest("POST", collectorURL, bytes.NewReader(jsonData))
+	if err != nil {
+		debugLog.Printf("Error creating the HTTP request: %s\n", err)
+		return err
+	}
+
+	req.Header.Set("Content-Type", "application/json; charset=utf-8")
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		log.Fatalf("Error sending the HTTP request: %s\n", err)
+		return err
+	}
+	defer resp.Body.Close()
+
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		debugLog.Printf("Error reading the HTTP response body: %s\n", err)
+		return err
+	}
+	debugLog.Printf("Response: %s\n", respBody)
+	return nil
+}
+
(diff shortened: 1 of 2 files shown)


x/reporter/reporter.go
```diff
@@ -0,0 +1,94 @@
...
+	"io"
+	"log"
+	"math/rand"
+	"net/http"
+)
+
+var debugLog log.Logger = *log.New(io.Discard, "", 0)
+var httpClient = &http.Client{}
+
+func sendReport(record map[string]interface{}, collectorURL string) error {
```
fortuna, line 31: Please create a `Reporter` type that encapsulates the collector URL and sampling factors, so that a client can simply call `reporter.Report(report)`

fortuna (changes requested):

x/reporter/reporter.go
amircybersec: I made the changes to encapsulate the Record parameters. I was previously using the function name `sendReportRandomly` to be more descriptive but I switched to `Report` to be shorter.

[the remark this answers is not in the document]
amircybersec: I changed the type to `any` but it might be safer to define a more specific type. I guess we would also need to accept slice/array of JSON objects to be reported as a batch, e.g. `[{}, {}]`. In that case, the reporter would have to iterate through the slice and send each JSON record one by one.

x/reporter/reporter.go
```diff
@@ -0,0 +1,101 @@
...
+var debugLog log.Logger = *log.New(io.Discard, "", 0)
+var httpClient = &http.Client{}
+
+type Record struct {
+	collectorURL      string
+	successSampleRate float64
+	failureSampleRate float64
+	success           bool
+	record            interface{}
+}
```
fortuna, line 37: We need to separate what is a report and what's a reporter.

We will need to specify the exact format of a report, which can be a struct, but for now it's any, but you need a success field:
```go
type Report struct {
  any
  success bool
}
```

The reporter must be an interface, so that anyone can implement one:
```go
type Reporter struct {
    Report(r Report) error
}
```
Perhaps we need to rename the method to something else, like `Register`. I'm hesitant of using `Send` because some reporters may batch them and send later.

Then provide an implementation that sends to a remote collector. It should have the url, and rates in the implementation structure, so one can call `reporter.Report(report)`.

fortuna (changes requested):

x/reporter/reporter.go
amircybersec: Just wanted to clarify something; Right now I am separating URL and sample rates from the log record / JSON that include the connectivity test error log, and having that data in a higher level structure. From what I understand here, you'd prefer to have that information in the JSON report? such as in this example:
```
	jsonData := `{
		"meta": {
			"successSampleRate": 1.0,
			"failureSampleRate": 0.0
			"collectorURL": "https://example-collector.com/collect"
		},
		"proxy": "192.168.1.1:65000",
		"transport": "shadowsocks",
		"resolver": "8.8.8.8:53",
		"proto": "tcp",
		"prefix": "HTTP1/1",
		"time": "2021-01-01T00:00:00Z",
		"durationMs": 100,
		"error": {
			"operation": "read",
			"posixError": "ETIMEDOUT",
			"msg": "i/o timeout"
		}
	}`  
```

fortuna: There's a lot of work to be done there. It's probably better to discuss in a doc.

For now it can be something like this:

```json
{
		"sampling_fraction": 1.0,
		"proxy": "192.168.1.1:65000",
		"transport": "shadowsocks",
		"resolver": "8.8.8.8:53",
		"proto": "tcp",
		"prefix": "HTTP/1.1",
		"time": "2021-01-01T00:00:00Z",
		"durationMs": 100,
		"error": {
			"operation": "read",
			"posixError": "ETIMEDOUT",
			"msg": "i/o timeout"
		}
}
```

- No need to specify the collector URL.
- You only need to specify the sampling fraction for this specific request.

Later we will need to extract what is test-specific (resolver) and transport-specific (shadowsocks, prefix).
For the DNS test, we should include the test domain. Not that relevant for an encrypted transport, but it will help in other cases.

amircybersec: been working on a design doc for this and have updated the code to reflect some of the proposed design. I will share that shortly with you.

x/reporter/reporter.go
```diff
@@ -0,0 +1,211 @@
...
+	// Observations
+	Time       time.Time   `json:"time"`
+	DurationMs int64       `json:"durationMs"`
+	Error      interface{} `json:"error"`
+}
+
+type Report interface {
+	IsSuccess() bool
+	CanMarshal() bool
+}
```
fortuna, line 45:
```suggestion
type Report any

 type HasSuccess interface {
	IsSuccess() bool
}
```

x/reporter/reporter.go
```diff
@@ -0,0 +1,211 @@
...
+// Makes sure the Report type can be marshalled into JSON
+func (r ConnectivityReport) CanMarshal() bool {
+	_, err := json.Marshal(r)
+	if err != nil {
+		log.Printf("Error encoding JSON: %s\n", err)
+		return false
+	} else {
+		return true
+	}
+}
```
fortuna, line 65:
```suggestion
```

x/reporter/reporter.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+	} else {
+		fmt.Println("Report sent")
+		return nil
+	}
+}
+
+func (c *SamplingCollector) Collect(report Report) error {
+	var samplingRate float64
+	if report.IsSuccess() {
```
fortuna, line 167:
```suggestion
	hs, ok := report.(HasSuccess)
	if !ok {
	    return error.New("report for SamplingCollector must implement HasSuccess")
	}
	if hs.IsSuccess() {
```

x/reporter/reporter.go
```diff
@@ -0,0 +1,211 @@
...
+func (r ConnectivityReport) CanMarshal() bool {
+	_, err := json.Marshal(r)
+	if err != nil {
+		log.Printf("Error encoding JSON: %s\n", err)
+		return false
+	} else {
+		return true
+	}
+}
+
```
fortuna, line 66: TODO: `WriteCollector` that takes a `io.Writer`

x/reporter/reporter.go
```diff
@@ -0,0 +1,211 @@
...
+type CollectorTarget struct {
+	collector Collector
+	priority  int
+	maxRetry  int
+}
+
+// TODO: implement a rotating collector
+type RotatingCollector struct {
```
fortuna, line 88: Separate use cases. Perhaps:
- `FallbackCollector` that tries them in order, until success.
- `MultiCollector` that collects on all of them. No need for priorities. Similar to `io.MultiWriter`.

x/reporter/reporter.go
```diff
@@ -0,0 +1,211 @@
...
+type SamplingCollector struct {
+	collector       Collector
+	successFraction float64
+	failureFraction float64
+}
+
+type CollectorTarget struct {
+	collector Collector
+	priority  int
+	maxRetry  int
```
fortuna, line 84: Create a `RetryCollector(collector, numRetries)` instead.

We will need to define an error type that allows the RetryCollector to determine whether the error is due to a bad request or not (like 4xx vx 5xx in HTTP), because bad requests are not retriable.

x/reporter/reporter.go
```diff
@@ -0,0 +1,211 @@
...
+	if err != nil {
+		log.Printf("Error encoding JSON: %s\n", err)
+		return false
+	} else {
+		return true
+	}
+}
+
+type Collector interface {
+	Collect(Report) error
```
fortuna, line 68: Let's pass a context to allow for timeouts and cancellations.
```suggestion
	Collect(context.Context, Report) error
```

x/reporter/reporter.go
fortuna: The error should be a `BadRequest`, so the retrier knows it shouldn't retry.

x/reporter/reporter.go
```diff
@@ -0,0 +1,211 @@
...
+func sendReport(jsonData []byte, remote *url.URL) (int, error) {
+	// TODO: return status code of HTTP response
+	req, err := http.NewRequest("POST", remote.String(), bytes.NewReader(jsonData))
+	if err != nil {
+		debugLog.Printf("Error creating the HTTP request: %s\n", err)
+		return 0, err
+	}
+
+	req.Header.Set("Content-Type", "application/json; charset=utf-8")
+	resp, err := httpClient.Do(req)
```
fortuna, line 196: Make sure you pass the `Context` here

fortuna (changes requested):

x/reporter/reporter.go
amircybersec: @fortuna we now have implementation for both `FallbackCollector` and `MultiCollector`

x/reporter/reporter.go
amircybersec: `BadRequestErr` is now defined for all HTTP error codes >= 400. I am passing the actual error code inside  `BadRequestErr`

fortuna (changes requested):

x/reporter/reporter.go
```diff
@@ -0,0 +1,216 @@
...
+	"math/rand"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+var debugLog log.Logger = *log.New(io.Discard, "", 0)
+var httpClient = &http.Client{}
+
+type BadRequestErr struct {
```
fortuna, line 34: It seems the standard library uses the `Err` prefix for error constants and the `Error` suffix for type, so we should use `BadRequestError` here.

PR #133, opened 2023-11-09
feat(transport): general TLS ClientHello fragmentation StreamDialer

jyyi1:
We introduce a new `StreamDialer` designed for TLS ClientHello fragmentation, as discussed in [*Circumventing the GFW with TLS Record Fragmentation*](https://upb-syssec.github.io/blog/2023/record-fragmentation/#tls-record-fragmentation). This general-purpose dialer allows for implementing custom splitting logic, enabling diverse fragmentation strategies.

Here is a screenshot that demonstrates the impact of using this StreamDialer to always split at the first 5 bytes of the Client Hello packet:

image


transport/tlsfrag/buffer.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+	return b.data[tlsRecordHeaderSize:b.len]
+}
+
+// Split fragments the Client Hello packet into two TLS records at the specified 0-based splitBytes:
+// [:splitBytes] and [splitBytes:]. Any necessary headers will be added to this buffer.
+//
+// If the packet has already be splitted before, a non-nil error and returned.
+// If the split index is ≤ 0 or ≥ the total length, do nothing.
+func (b *clientHelloBuffer) Split(splitBytes int) error {
```
fortuna, line 215: As discussed offline, let's separate the record buffering and the split logic, so it's easier to understand and modify.

transport/tlsfrag/buffer.go
```diff
@@ -0,0 +1,307 @@
...
+func recordLen(hdr []byte) uint16 {
+	return binary.BigEndian.Uint16(hdr[3:])
+}
+
+func isValidRecordLenForHandshake(len uint16) bool {
+	return 0 < len && len <= tlsMaxRecordLen
+}
+
+func putTLSClientHelloHeader(hdr []byte, recordLen uint16) {
+	_ = hdr[4] // bounds check to guarantee safety of writes below
```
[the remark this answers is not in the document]
jyyi1: https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/encoding/binary/binary.go;l=102

transport/tlsfrag/buffer.go
```diff
@@ -0,0 +1,307 @@
...
+	tlsRecordHeaderSize            = 5 // the minimum size that contains the entire header
+	tlsTypeHandshake               = 22
+	tlsMaxRecordLen                = 1 << 14
+)
+
+// errInvalidTLSClientHello is the error used when the data received is not a valid TLS Client Hello.
+// Please use [errors.Is] to compare the returned err object with this instance.
+var errInvalidTLSClientHello = errors.New("not a valid TLS Client Hello packet")
+
+func isTLSRecordTypeHandshake(hdr []byte) bool {
```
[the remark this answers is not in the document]
jyyi1: We can use the `_ = hdr[x]` to check instead of returning an `error` object, similar to the `binary` package implementation mentioned below.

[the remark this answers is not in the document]
jyyi1: Right, I'm aware of that. `panic` is the expected behavior. The early check is to prevent putting garbage data into the buffer if it's not big enough.

transport/tlsfrag/buffer.go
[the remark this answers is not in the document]
jyyi1: Hmm, not as I expect. The expected behavior is:

```
if len(hdr) < 4 then
  panic without writing a single byte of data
else
  write data
```

transport/tlsfrag/writer.go
```diff
@@ -0,0 +1,179 @@
...
+	split := w.frag(content)
+	if split <= 0 || split >= len(content) {
+		w.copyBufToRecords()
+		return
+	}
+
+	header := make([]byte, recordHeaderLen)
+	copy(header, w.buf.Bytes())
+
+	w.rcds = bytes.NewBuffer(make([]byte, 0, w.buf.Len()+recordHeaderLen))
```
fortuna, line 158: Please remove this allocation and copy. You can write directly from the record buffer:
1. update record len for head
2. shift tail and add a header for it with proper length
3. write the full buffer

This requires the buffer to have recordLength + 2 * recordHeaderLen bytes

transport/tlsfrag/writer.go
```diff
@@ -0,0 +1,179 @@
...
+// clientHelloFragWriter intercepts the initial TLS Client Hello record and splits it into two TLS records based on the
+// return value of frag function. These fragmented records are then written to the base [io.Writer]. Subsequent packets
+// are not modified and are directly transmitted through the base [io.Writer].
+type clientHelloFragWriter struct {
+	base io.Writer
+	done bool // indicates all splitted rcds have been already written to base
+	frag FragFunc
+
+	buf  *clientHelloBuffer // the buffer containing and parsing a TLS Client Hello record
+	rcds *bytes.Buffer      // the buffer containing splitted records what will be written to base
```
fortuna, line 32: I believe this holds only one record?
```suggestion
	record *bytes.Buffer      // the buffer containing splitted records what will be written to base
```

fortuna (changes requested):

transport/tlsfrag/buffer.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+var _ io.Writer = (*clientHelloBuffer)(nil)
+var _ io.ReaderFrom = (*clientHelloBuffer)(nil)
+
+// newClientHelloBuffer creates and initializes a new buffer to receive a TLS Client Hello packet.
+func newClientHelloBuffer() *clientHelloBuffer {
+	// Allocate the 5 bytes header first, and then reallocate it to contain the entire packet later
+	return &clientHelloBuffer{
+		data:   make([]byte, recordHeaderLen),
```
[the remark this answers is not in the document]
jyyi1: I tried that before, and it will cause additional buffer allocations in `ReadFrom`.

Think of the scenario reading from `r io.Reader`, you need to call `r.Read(p)`, and `p` cannot be empty. If you use `r.Read(data[:5])`, the `data` will be expanded to 5 bytes immediately, what if `r` doesn't have that many bytes available (like only 3 bytes is in `r`), when `len(b.data) == 5` but actually there are only 3 valid bytes read.

transport/tlsfrag/buffer.go
```diff
@@ -0,0 +1,153 @@
...
+	"io"
+)
+
+var (
+	// errTLSClientHelloFullyReceived is returned when a full TLS Client Hello has been received and no
+	// more data can be pushed to the buffer.
+	errTLSClientHelloFullyReceived = errors.New("already received a complete TLS Client Hello packet")
+
+	// errInvalidTLSClientHello is the error used when the data received is not a valid TLS Client Hello.
+	// Please use [errors.Is] to compare the returned err object with this instance.
```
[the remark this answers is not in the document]
jyyi1: Depending on are we going to return a nested error like `fmt.Errorf(..., interestingErr)` in the code. For example, I didn't add that for `errTLSClientHelloFullyReceived` because we will directly return this error. But if `errors.Is` is the principle in Go, we can omit that.

transport/tlsfrag/buffer.go
```diff
@@ -0,0 +1,153 @@
...
+		m := copy(b.data[b.len:], p)
+		n += m
+		b.len += m
+		p = p[m:]
+
+		if b.len == recordHeaderLen {
+			if err = b.validateTLSClientHello(); err != nil {
+				return
+			}
+			buf := make([]byte, recordHeaderLen+getMsgLen(b.data))
```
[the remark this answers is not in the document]
jyyi1: I tried it before, but `bytes.Buffer` has a limitation when `ReadFrom` a `r io.Reader`. Because it implements `ReadFrom`, but in the implementation it is allocating additional buffers (with minimum size 512 bytes) instead of using existing buffer until `cap()`: https://cs.opensource.google/go/go/+/refs/tags/go1.21.3:src/bytes/buffer.go;l=209. This might over read from `r`, meaning reading more bytes than we needed.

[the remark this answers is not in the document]
jyyi1: We can implement that. But just note that it won't completely solve the timeout or write alignment issue:

* **Timeout** user can `Write` 4 valid bytes to us, and then `Read`, we will wait the 5th byte forever because we need to that to determine the total size.
* **Write Alignment** To solve it, we need additional allocation: user `split a 15 bytes (containing 10 bytes message) packet into three writes `Write(7)` `Write(5)` `Write(3)`, suppose we have a fixed 2 bytes split point, we will forward: `Write(7)` (with modified the new msgsize = 2), `Write(5)` (the new header with msgsize=8), `Write(5)` and than `Write(3)`. There is an additional `Write(5)` there. To merge it with the following content, we need to allocate a 10 bytes buffer to put both the header and the content.

transport/tlsfrag/buffer.go
```diff
@@ -0,0 +1,153 @@
...
+// If this buffer still requires more data to build a complete TLS Client Hello message, it returns nil error.
+//
+// You can call ReadFrom multiple times if r doesn't provide enough data to build a complete Client Hello packet.
+func (b *clientHelloBuffer) ReadFrom(r io.Reader) (n int64, err error) {
+	if !b.valid {
+		return 0, errInvalidTLSClientHello
+	}
+
+	for b.len < len(b.data) && err == nil {
+		m, e := r.Read(b.data[b.len:])
```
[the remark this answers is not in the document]
jyyi1: I'm doing it in a `for`.

transport/tlsfrag/tls.go
```diff
@@ -0,0 +1,95 @@
...
+//	|   Record    |
+//	|   Length    |
+//	+-------------+ 5
+//	|   Message   |
+//	|    Data     |
+//	|     ...     |
+//	+-------------+ Message Length + 5
+//
+//	RecordType := invalid(0) | handshake(22) | application_data(23) | ...
+//	Protocol Version (deprecated) := 0x0301 ("TLS 1.0") | 0x0303 ("TLS 1.2" & "TLS 1.3") | 0x0302 ("TLS 1.1")
```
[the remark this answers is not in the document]
jyyi1: You mean in alphabetical order? Cuz I sorted them in popularity order. 😄 Will do.

transport/tlsfrag/tls.go
```diff
@@ -0,0 +1,95 @@
...
+//	0 < Message Length (of handshake)        ≤ 2^14
+//	0 ≤ Message Length (of application_data) ≤ 2^14
+//
+// [RFC 8446]: https://datatracker.ietf.org/doc/html/rfc8446#section-5.1
+const (
+	recordHeaderLen = 5
+	maxMsgLen       = 1 << 14
```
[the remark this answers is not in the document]
jyyi1: How about `maxPayloadLen`?

transport/tlsfrag/tls.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+// getRecordType gets the TLS record type from the TLS header hdr[0]. This function will panic if len(hdr) < 1.
+func getRecordType(hdr []byte) recordType {
+	return recordType(hdr[0])
+}
+
+// getTLSVersion gets the TLS version from the TLS header hdr[1:3]. This function will panic if len(hdr) < 3.
+func getTLSVersion(hdr []byte) tlsVersion {
+	return tlsVersion(binary.BigEndian.Uint16(hdr[1:]))
```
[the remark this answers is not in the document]
jyyi1: Not necessary for logic, `Uint16` will only read 2 bytes, but for readability, yes.

transport/tlsfrag/writer.go
jyyi1: Yeah, but there is a tradeoff here, we either need to expose implementation details in `buffer.go` (with an extra 5 bytes padding that is not used anywhere in `buffer.go`); if we'd like to keep the type doing what it should do, we need to allocate additional space.

transport/tlsfrag/writer.go
```diff
@@ -0,0 +1,179 @@
...
+	return fw, nil
+}
+
+// Write implements io.Writer.Write. It attempts to split the data received in the first one or more Write call(s)
+// into two TLS records if the data corresponds to a TLS Client Hello record.
+func (w *clientHelloFragWriter) Write(p []byte) (n int, err error) {
+	if w.done {
+		return w.base.Write(p)
+	}
+	if w.rcds != nil {
```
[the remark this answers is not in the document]
jyyi1: This is when the first split Write fails, and the caller just ignores the error and issues another `Write`:

```go
Write() // splitted to two records, tried to Write the records but failed, only part of it is flushed to the base Writer
Write() // need to flush the remaining part in records above, and then Write all p to the base Writer
```

transport/tlsfrag/writer.go
jyyi1: It contains two splitted records.

transport/tlsfrag/writer.go
```diff
@@ -0,0 +1,179 @@
...
+	"errors"
+	"io"
+)
+
+// clientHelloFragWriter intercepts the initial TLS Client Hello record and splits it into two TLS records based on the
+// return value of frag function. These fragmented records are then written to the base [io.Writer]. Subsequent packets
+// are not modified and are directly transmitted through the base [io.Writer].
+type clientHelloFragWriter struct {
+	base io.Writer
+	done bool // indicates all splitted rcds have been already written to base
```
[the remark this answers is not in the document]
jyyi1: Two spaces will be reduced to one by `go fmt`.

transport/tlsfrag/buffer.go
```diff
@@ -0,0 +1,153 @@
...
+	toRead int    // the number of bytes to read next, e.g. 5 -> len(content)
+}
+
+var _ io.Writer = (*clientHelloBuffer)(nil)
+var _ io.ReaderFrom = (*clientHelloBuffer)(nil)
+
+// newClientHelloBuffer creates and initializes a new buffer to receive a TLS Client Hello packet.
+func newClientHelloBuffer() *clientHelloBuffer {
+	// Allocate the 5 bytes header first, and then reallocate it to contain the entire packet later
+	return &clientHelloBuffer{
```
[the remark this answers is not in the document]
jyyi1: This will create a `tlsRecordHeader` instead of taking it because it will buffer until a valid header is read as well.

transport/socks5/socks5.go
```diff
@@ -109,6 +110,6 @@ func appendSOCKS5Address(b []byte, address string) ([]byte, error) {
 		b = append(b, byte(len(host)))
 		b = append(b, host...)
 	}
-	b = append(b, byte(portNum>>8), byte(portNum))
+	b = binary.BigEndian.AppendUint16(b, uint16(portNum))
```
jyyi1, line 113: This is to resolve the CodeQL error: https://github.com/Jigsaw-Code/outline-sdk/pull/133/checks?check_run_id=18912663629

(Although the standard library `AppendUint16` is using the same code to append the data)

transport/tlsfrag/buffer.go
```diff
@@ -0,0 +1,134 @@
...
+	if err == io.EOF {
+		err = nil
+	}
+	if len(b.data) == cap(b.data)-b.padding {
+		err = errors.Join(err, errTLSClientHelloFullyReceived)
+	}
+	return
+}
+
+func (b *clientHelloBuffer) validateTLSClientHello() (tlsRecordHeader, error) {
```
fortuna, line 119: Can you make this a standalone function by passing `b.data` as input? Clear inputs are a lot easier to understand. Otherwise it's hard to tell it doesn't use any other data.
You can set `b.valid` in the caller.

transport/tlsfrag/buffer.go
```diff
@@ -0,0 +1,134 @@
...
+// received, or r returns EOF or error. It returns the number of bytes read. Any error except EOF encountered during
+// the read is also returned.
+//
+// If this buffer completes a valid TLS Client Hello, it returns errTLSClientHelloFullyReceived.
+// If an invalid TLS Client Hello message is detected, it returns the error errInvalidTLSClientHello.
+// If this buffer still requires more data to build a complete TLS Client Hello message, it returns nil error.
+//
+// You can call ReadFrom multiple times if r doesn't provide enough data to build a complete Client Hello packet.
+func (b *clientHelloBuffer) ReadFrom(r io.Reader) (n int64, err error) {
+	if !b.valid {
```
fortuna, line 85: This can move to after we have the header.

transport/tlsfrag/buffer.go
```diff
@@ -0,0 +1,134 @@
...
+var _ io.Writer = (*clientHelloBuffer)(nil)
+var _ io.ReaderFrom = (*clientHelloBuffer)(nil)
+
+// newClientHelloBuffer creates and initializes a new buffer to receive a TLS Client Hello packet.
+func newClientHelloBuffer() *clientHelloBuffer {
+	// Allocate the 5 bytes header first, and then reallocate it to contain the entire packet later
+	return &clientHelloBuffer{
+		data:    make([]byte, 0, recordHeaderLen),
+		padding: 0,
+		valid:   true,
```
fortuna, line 51: You could also define this as `validationStatus error`, so that you can keep returning the same error.

transport/tlsfrag/writer.go
```diff
@@ -0,0 +1,177 @@
...
+	"errors"
+	"io"
+)
+
+// clientHelloFragWriter intercepts the initial TLS Client Hello record and splits it into two TLS records based on the
+// return value of frag function. These fragmented records are then written to the base [io.Writer]. Subsequent packets
+// are not modified and are directly transmitted through the base [io.Writer].
+type clientHelloFragWriter struct {
+	base io.Writer
+	done bool // Indicates all splitted rcds have been already written to base
```
fortuna, line 28: I think these comments are more readable if they are places on the line before the field.

transport/tlsfrag/writer.go
```diff
@@ -0,0 +1,177 @@
...
+func (w *clientHelloFragWriter) splitHelloBufToRecord() {
+	received := w.helloBuf.Bytes()
+	content := received[recordHeaderLen:]
+	split := w.frag(content)
+	if split <= 0 || split >= len(content) {
+		w.copyHelloBufToRecord()
+		return
+	}
+
+	// received: | <== header (5) ==> | <== split ==> | <== len(content)-split ==> |  ... cap with padding (5) ... |
```
fortuna, line 151: Not quite readable on a narrower screen:
image

transport/tlsfrag/writer.go
```diff
@@ -0,0 +1,177 @@
...
+
+// copyHelloBufToRecord copies w.helloBuf into w.record without allocations.
+func (w *clientHelloFragWriter) copyHelloBufToRecord() {
+	w.record = bytes.NewBuffer(w.helloBuf.Bytes())
+	w.helloBuf = nil // allows the GC to recycle the memory
+}
+
+// splitHelloBufToRecord splits w.helloBuf into two records and put them into w.record without allocations.
+func (w *clientHelloFragWriter) splitHelloBufToRecord() {
+	received := w.helloBuf.Bytes()
```
fortuna, line 143: There are some ambiguous names here. Perhaps call them originalRecord, headRecord and tailRecord to be clearer?:
```suggestion
	originalRecord := w.helloBuf.Bytes()
```

transport/tlsfrag/writer.go
```diff
@@ -0,0 +1,177 @@
...
+	content := received[recordHeaderLen:]
+	split := w.frag(content)
```
fortuna, line 145: Perhaps call the fragments head and tail? Then we can have:
```suggestion
	headLen := w.frag(content)
```

transport/tlsfrag/writer.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// received: | <== header (5) ==> | <== split ==> | <== len(content)-split ==> |  ... cap with padding (5) ... |
+	//                                                 \                            \
+	//                                                  +-----------------+          +-----------------+
+	//                                                                     \                            \
+	// splitted: | <== header (5) ==> | <== split ==> | <== header2 (5) ==> | <== len(content)-split ==> |
+	splitted := received[:len(received)+recordHeaderLen]
+	hdr1 := tlsRecordHeaderFromRawBytes(splitted[:recordHeaderLen])
+	hdr2 := tlsRecordHeaderFromRawBytes(splitted[recordHeaderLen+split : recordHeaderLen*2+split])
+	recvContent2 := splitted[recordHeaderLen+split : len(received)]
```
fortuna, line 159: We need some comments
```suggestion
        // Shift tail fragment to make space for record header.
	recvContent2 := splitted[recordHeaderLen+split : len(received)]
```

transport/tlsfrag/writer.go
```diff
@@ -0,0 +1,177 @@
...
+	content2 := splitted[recordHeaderLen*2+split:]
+	copy(content2, recvContent2)
+	copy(hdr2, hdr1)
+	hdr1.SetPayloadLen(uint16(split))
```
fortuna, line 163: This should move up. Do the first fragment, then the second.

transport/tlsfrag/writer.go
```diff
@@ -0,0 +1,177 @@
...
+	hdr2.SetPayloadLen(uint16(len(content) - split))
```
fortuna, line 164:
```suggestion
        // Insert header for second fragment.
	copy(hdr2, hdr1)
	hdr2.SetPayloadLen(uint16(len(content) - split))
```

transport/tlsfrag/writer.go
```diff
@@ -0,0 +1,177 @@
...
+	w.record = bytes.NewBuffer(splitted)
```
fortuna, line 165: Why don't you write to the base writer here?

transport/tlsfrag/writer.go
```diff
@@ -0,0 +1,177 @@
...
+
+	m, e := w.baseRF.ReadFrom(r)
+	n += m
+	err = e
+	return
+}
+
+// copyHelloBufToRecord copies w.helloBuf into w.record without allocations.
+func (w *clientHelloFragWriter) copyHelloBufToRecord() {
+	w.record = bytes.NewBuffer(w.helloBuf.Bytes())
```
fortuna, line 137: Why don't you just write to the base writer here?

fortuna (changes requested):

transport/tlsfrag/writer.go
jyyi1: Hmm, indeed. Let me remove it then.

transport/tlsfrag/writer.go
jyyi1: A single `Write` might fail, so you still need to track the remaining bytes somewhere in `b`, and you need to `Write` them again in the following `Write`s or `ReadFrom`s issued by the caller.

I think it makes more sense to separate the "flush bytes to base" logic from the `split`.

transport/tlsfrag/writer.go
jyyi1: Same as above. ☝️

transport/tlsfrag/buffer.go
```diff
@@ -0,0 +1,134 @@
...
+func (b *clientHelloBuffer) ReadFrom(r io.Reader) (n int64, err error) {
+	if !b.valid {
+		return 0, errInvalidTLSClientHello
+	}
+
+	for len(b.data) < cap(b.data)-b.padding && err == nil {
+		m, e := r.Read(b.data[len(b.data) : cap(b.data)-b.padding])
+		b.data = b.data[:len(b.data)+m]
+		n += int64(m)
+		err = e
```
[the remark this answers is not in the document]
jyyi1: Because according to the doc https://pkg.go.dev/io#Reader:

> Callers should always process the n > 0 bytes returned before considering the error err.

In our case, "processing" means determine whether we have read a full header or a full record. After that, we will jump out of the loop when `err != nil` (see the condition in `for` loop).

transport/tlsfrag/buffer.go
```diff
@@ -0,0 +1,134 @@
...
+
+		if len(b.data) == recordHeaderLen {
+			hdr, e := b.validateTLSClientHello()
+			if e != nil {
+				if err == io.EOF {
+					err = nil
+				}
+				err = errors.Join(err, e)
```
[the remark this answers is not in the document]
jyyi1: Make sense if the only user is `writer.go`. I can keep the logic simpler for now.

transport/tlsfrag/buffer.go
```diff
@@ -0,0 +1,134 @@
...
+			b.data = append(buf, b.data...)
+			b.padding = recordHeaderLen
+		}
+	}
+
+	if err == io.EOF {
+		err = nil
+	}
+	if len(b.data) == cap(b.data)-b.padding {
+		err = errors.Join(err, errTLSClientHelloFullyReceived)
```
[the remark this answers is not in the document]
jyyi1: Yes, we don't have consumers of other errors for now, we can simplify the logic here.

transport/tlsfrag/tls.go
```diff
@@ -0,0 +1,102 @@
...
+	recordTypeHandshake recordType = 22
+
+	versionTLS10 tlsVersion = 0x0301
+	versionTLS11 tlsVersion = 0x0302
+	versionTLS12 tlsVersion = 0x0303
+	versionTLS13 tlsVersion = 0x0304
+)
+
+// tlsRecordHeaderFromRawBytes creates a new tlsRecordHeader from raw. This function will panic if len(raw) < 5.
+func tlsRecordHeaderFromRawBytes(raw []byte) tlsRecordHeader {
```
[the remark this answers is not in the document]
jyyi1: I've tried but did not figure out a simpler way of handling updates. Because `[5]byte` is an array, so it will be value-copied, for example after `g := tlsRecordHeader(b.data)`, `g` is a copy of `b.data` instead of a reference, so any updates to `g` won't be reflected to `b.data`, we have to introduced another helper method `func (h tlsRecordHeader) PutToSlice(data []byte)` to copy all bytes from `h` to `data`, that's more complicated if we simply define `tlsRecordHeader` as a slice.

transport/tlsfrag/buffer.go
```diff
@@ -0,0 +1,134 @@
...
+// If this buffer still requires more data to build a complete TLS Client Hello message, it returns nil error.
+//
+// You can call ReadFrom multiple times if r doesn't provide enough data to build a complete Client Hello packet.
+func (b *clientHelloBuffer) ReadFrom(r io.Reader) (n int64, err error) {
+	if !b.valid {
+		return 0, errInvalidTLSClientHello
+	}
+
+	for len(b.data) < cap(b.data)-b.padding && err == nil {
+		m, e := r.Read(b.data[len(b.data) : cap(b.data)-b.padding])
```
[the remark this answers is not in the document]
jyyi1: Refactored it into 3 stages: appending header; validation; appending payload.

transport/tlsfrag/writer.go
jyyi1: Reduced it to 67 characters, should be readable now.

PR #166, opened 2024-01-18
Update iOS WebView interception instructions

daniellacosse:
Let me know if you'd like more detail.


fortuna (approved):
Thanks for improving the doc!

PR #189, opened 2024-02-25
feat: add User/Pass Authentication to Sock5 Dialer with Tests

amircybersec:
This PR adds username/password authentication to Socks5 dialer. Changes to config package will come in a separate PR to avoid breaking build checks. The config URL for socks5 with authentication would like this:

```
socks5://username:password@hostname:port
```

For testing I am currently using [this](https://github.com/armon/go-socks5) socks5 server implementation. 

TODO
- [x] Add tests for authentication 
- [x] Connect in a single round trip


amircybersec:
@fortuna I will create a separate PR for config package changes to accept config URL with username/password for socks5 after this PR is merged otherwise the config package PR breaks the build.

transport/socks5/stream_dialer.go
```diff
@@ -55,78 +86,144 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+
+	if c.credentials == nil {
+		// Method selection part: VER = 5, NMETHODS = 1, METHODS = 0 (no auth)
+		// +----+----------+----------+
+		// |VER | NMETHODS | METHODS  |
+		// +----+----------+----------+
+		// | 1  |    1     | 1 to 255 |
+		// +----+----------+----------+
+		header := [3 + 3 + 256 + 2]byte{}
+		buffer = append(header[:0], 5, 1, 0)
```
fortuna, line 102: Please clarify the comments:
```suggestion
		// Buffer large enough for method and connect requests with a domain name address.
		header := [3 + (3 + 256 + 2)]byte{}
		// Method selection part: VER = 5, NMETHODS = 1, METHODS = 0 (no auth)
		// +----+----------+----------+
		// |VER | NMETHODS | METHODS  |
		// +----+----------+----------+
		// | 1  |    1     | 1 to 255 |
		// +----+----------+----------+
		buffer = append(header[:0], 5, 1, 0)
```

I also think it's a good idea to add parenthesis to clarify the blocks.

transport/socks5/stream_dialer.go
```diff
@@ -55,78 +86,144 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+	} else {
+		// https://datatracker.ietf.org/doc/html/rfc1929
+		// Method selection part: VER = 5, NMETHODS = 1, METHODS = 2 (username/password)
+		header := [3 + 3 + 255 + 255 + 3 + 256 + 2]byte{}
+		buffer = append(header[:0], 5, 1, 2)
```
fortuna, line 107: Attach the comment to the corresponding line:
```suggestion
		header := [3 + 3 + 255 + 255 + 3 + 256 + 2]byte{}
		// Method selection part: VER = 5, NMETHODS = 1, METHODS = 2 (username/password)
		buffer = append(header[:0], 5, 1, 2)
```

transport/socks5/stream_dialer.go
```diff
@@ -55,78 +86,144 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+		// +----+------+----------+------+----------+
+		// |VER | ULEN |  UNAME   | PLEN |  PASSWD  |
+		// +----+------+----------+------+----------+
+		// | 1  |  1   | 1 to 255 |  1   | 1 to 255 |
+		// +----+------+----------+------+----------+
+		buffer = append(buffer, 1) // Auth version
+		buffer = append(buffer, byte(len(c.credentials.username)))
+		buffer = append(buffer, c.credentials.username...)
+		buffer = append(buffer, byte(len(c.credentials.password)))
+		buffer = append(buffer, c.credentials.password...)
```
fortuna, line 119: Less verbose and more readable:
```suggestion
		buffer = append(buffer,
			1,
			byte(len(c.credentials.username)),
			c.credentials.username...,
			byte(len(c.credentials.password)),
			c.credentials.password...
		)
```

transport/socks5/stream_dialer.go
```diff
@@ -55,78 +86,144 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+		// |VER | NMETHODS | METHODS  |
+		// +----+----------+----------+
+		// | 1  |    1     | 1 to 255 |
+		// +----+----------+----------+
+		header := [3 + 3 + 256 + 2]byte{}
+		buffer = append(header[:0], 5, 1, 0)
+	} else {
+		// https://datatracker.ietf.org/doc/html/rfc1929
+		// Method selection part: VER = 5, NMETHODS = 1, METHODS = 2 (username/password)
+		header := [3 + 3 + 255 + 255 + 3 + 256 + 2]byte{}
```
fortuna, line 106:
```suggestion
		// Buffer large enough for method, credentials, and connect requests with a domain name address.
		header := [3 + (1 + 1 + 255 + 1 + 255) + (3 + 256 + 2)]byte{}
```

transport/socks5/stream_dialer.go
```diff
@@ -55,78 +86,144 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
-	// Method request:
-	// VER = 5, NMETHODS = 1, METHODS = 0 (no auth)
-	b := append(header[:0], 5, 1, 0)
-
-	// Connect request:
-	// VER = 5, CMD = 1 (connect), RSV = 0
-	b = append(b, 5, 1, 0)
-	// Destination address Address (ATYP, DST.ADDR, DST.PORT)
-	b, err = appendSOCKS5Address(b, remoteAddr)
+	// Connect request part: VER = 5, CMD = 1 (connect), RSV = 0, DST.ADDR, DST.PORT
```
fortuna, line 122: Line break for readability:
```suggestion
	// Connect request:
	// VER = 5, CMD = 1 (connect), RSV = 0, DST.ADDR, DST.PORT
```

transport/socks5/stream_dialer.go
```diff
@@ -55,78 +86,144 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+	// +----+-----+-------+------+----------+----------+
+	// |VER | CMD |  RSV  | ATYP | DST.ADDR | DST.PORT |
+	// +----+-----+-------+------+----------+----------+
+	// | 1  |  1  | X'00' |  1   | Variable |    2     |
+	// +----+-----+-------+------+----------+----------+
+	connectRequest, err := appendSOCKS5Address([]byte{5, 1, 0}, remoteAddr)
```
fortuna, line 128: This is allocating a new buffer for the connect request. 
To minimize the allocations, please add to the buffer directly like we were doing before.

transport/socks5/stream_dialer.go
```diff
@@ -55,78 +86,144 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to create SOCKS5 address: %w", err)
 	}
+	buffer = append(buffer, connectRequest...)
 
-	// We merge the method and connect requests because we send a single authentication
```
fortuna, line 75: This comment is important because it clarifies why we do the extra effort to do one write. Please keep it. It needs to be updated with "This eliminates a roundtrip, two if there is authentication."

transport/socks5/stream_dialer.go
```diff
@@ -55,78 +86,144 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+	// Sending the combined request
+	_, err = proxyConn.Write(buffer)
 	if err != nil {
-		return nil, fmt.Errorf("failed to write SOCKS5 request: %w", err)
+		return nil, fmt.Errorf("failed to write combined SOCKS5 request: %w", err)
 	}
 
-	// Read method response (VER, METHOD).
-	if _, err = io.ReadFull(proxyConn, header[:2]); err != nil {
+	// Read several response parts in one go, to avoid an unnecessary roundtrip.
```
fortuna, line 140: Remove the comment. We are not really reading the response in one go. We are reading piece by piece. In fact, the server may reply with separate packets.

transport/socks5/stream_dialer.go
```diff
@@ -55,78 +86,144 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+	var methodResponse [2]byte
+	if _, err = io.ReadFull(proxyConn, methodResponse[:]); err != nil {
 		return nil, fmt.Errorf("failed to read method server response")
 	}
-	if header[0] != 5 {
-		return nil, fmt.Errorf("invalid protocol version %v. Expected 5", header[0])
+	if methodResponse[0] != 5 {
+		return nil, fmt.Errorf("invalid protocol version %v. Expected 5", methodResponse[0])
+	}
+	if methodResponse[1] == 2 {
```
fortuna, line 154: Let's define constants for the method values.

fortuna, line 154: Use a switch on the `methodResponse[1]` instead, so it's clear the different options. The default should fail

transport/socks5/stream_dialer.go
```diff
@@ -55,78 +86,144 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
-	if _, err = io.ReadFull(proxyConn, header[:2]); err != nil {
+	// Read several response parts in one go, to avoid an unnecessary roundtrip.
+	// 1. Read method response (VER, METHOD).
+	// +----+--------+
+	// |VER | METHOD |
+	// +----+--------+
+	// | 1  |   1    |
+	// +----+--------+
+	var methodResponse [2]byte
+	if _, err = io.ReadFull(proxyConn, methodResponse[:]); err != nil {
```
fortuna, line 148: I believe this escapes to the heap (you can check), which means methodResponse will be allocated on the heap.
We should minimize allocations because they make the garbage collector trigger more frequently, and it causes issues on iOS where the VPNExtension may get killed for excess memory.

Reuse the header instead. You can do:
```go
methodresponse := header[:2]
```

transport/socks5/stream_dialer.go
```diff
@@ -55,78 +86,144 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+	}
+	if methodResponse[1] == 2 {
+		// 2. Read sub-negotiation version and status
+		// VER = 1, STATUS = 0
+		// +----+--------+
+		// |VER | STATUS |
+		// +----+--------+
+		// | 1  |   1    |
+		// +----+--------+
+		var subNegotiation [2]byte
```
fortuna, line 162: Reuse buffer

transport/socks5/stream_dialer.go
```diff
@@ -55,78 +86,144 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+	if _, err = io.ReadFull(proxyConn, methodResponse[:]); err != nil {
 		return nil, fmt.Errorf("failed to read method server response")
 	}
-	if header[0] != 5 {
-		return nil, fmt.Errorf("invalid protocol version %v. Expected 5", header[0])
+	if methodResponse[0] != 5 {
+		return nil, fmt.Errorf("invalid protocol version %v. Expected 5", methodResponse[0])
+	}
+	if methodResponse[1] == 2 {
+		// 2. Read sub-negotiation version and status
```
fortuna, line 155: Sub-negotiation is a bad name.

Use authentication response/reply instead.

transport/socks5/stream_dialer.go
```diff
@@ -55,78 +86,144 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+		// |VER | STATUS |
+		// +----+--------+
+		// | 1  |   1    |
+		// +----+--------+
+		var subNegotiation [2]byte
+		if _, err = io.ReadFull(proxyConn, subNegotiation[:]); err != nil {
+			return nil, fmt.Errorf("failed to read sub-negotiation version and status: %w", err)
+		}
+		if subNegotiation[0] != 1 {
+			return nil, fmt.Errorf("unkown sub-negotioation version")
```
fortuna, line 167: Outli
```suggestion
			return nil, fmt.Errorf("unkown sub-negotioation version %v", subNegotiation[0])
```

transport/socks5/stream_dialer.go
```diff
@@ -55,78 +86,144 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+	// 3. Read connect response (VER, REP, RSV, ATYP, BND.ADDR, BND.PORT).
 	// See https://datatracker.ietf.org/doc/html/rfc1928#section-6.
-	if _, err = io.ReadFull(proxyConn, header[:4]); err != nil {
-		return nil, fmt.Errorf("failed to read connect server response")
+	// +----+-----+-------+------+----------+----------+
+	// |VER | REP |  RSV  | ATYP | BND.ADDR | BND.PORT |
+	// +----+-----+-------+------+----------+----------+
+	// | 1  |  1  | X'00' |  1   | Variable |    2     |
+	// +----+-----+-------+------+----------+----------+
+	var connectResponse [4]byte
```
fortuna, line 185: Reuse buffer

transport/socks5/stream_dialer.go
```diff
@@ -55,78 +86,144 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
-		return nil, fmt.Errorf("failed to read address in connect response: %w", err)
+	bndAddr := make([]byte, bndAddrLen)
+	if _, err = io.ReadFull(proxyConn, bndAddr); err != nil {
+		return nil, fmt.Errorf("failed to read bound address: %w", err)
 	}
 	// We also ignore the remote bound port number.
-	_, err = io.ReadFull(proxyConn, header[:2])
-	if err != nil {
-		return nil, fmt.Errorf("failed to read port number in connect response: %w", err)
+	// Read the port (2 bytes)
```
fortuna, line 223: Reverse the comment lines here.

fortuna (changes requested):

amircybersec:
@fortuna Thanks again for the feedback! I applied your comments; Could you please review this again for me? Btw, `header` still escapes to the heap based on escape analysis I did with `go build -gcflags "-m"`. Removing header from heap was not trivial and may not even be possible. I took the pre-allocated array approach in the original code to ensure more efficient memory allocation though.

amircybersec:
@fortuna also windows build failed again but I recall you mentioned it is due to some race condition. It sometimes pass for me and sometimes fail even though I am not making any breaking changes...

fortuna (changes requested):

transport/socks5/stream_dialer.go
```diff
@@ -55,80 +121,151 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+		// Method selection part: VER = 5, NMETHODS = 1, METHODS = 0 (no auth)
+		// +----+----------+----------+
+		// |VER | NMETHODS | METHODS  |
+		// +----+----------+----------+
+		// | 1  |    1     | 1 to 255 |
+		// +----+----------+----------+
+		b = append(buffer[:0], socksProtocolVer, numberOfAuthMethods, noAuthMethod)
+	} else {
+		// https://datatracker.ietf.org/doc/html/rfc1929
+		// Method selection part: VER = 5, NMETHODS = 1, METHODS = 2 (username/password)
```
fortuna, line 139:
```suggestion
		// Method selection part: VER = 5, NMETHODS = 1, METHOD = 2 (username/password)
```

transport/socks5/stream_dialer.go
```diff
@@ -55,80 +121,151 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+		b = append(buffer[:0], socksProtocolVer, numberOfAuthMethods, userPassAuthMethod)
```
fortuna, line 140: Not for this PR, but we could potentially add limited padding (up to 254 extra bytes) by creating a variable list of repeated methods.

transport/socks5/stream_dialer.go
```diff
@@ -55,80 +121,151 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
-	// Method request:
-	// VER = 5, NMETHODS = 1, METHODS = 0 (no auth)
-	b := append(header[:0], 5, 1, 0)
+		// Authentication part: VER = 1, ULEN = 1, UNAME = 1~255, PLEN = 1, PASSWD = 1~255
+		// +----+------+----------+------+----------+
+		// |VER | ULEN |  UNAME   | PLEN |  PASSWD  |
+		// +----+------+----------+------+----------+
+		// | 1  |  1   | 1 to 255 |  1   | 1 to 255 |
+		// +----+------+----------+------+----------+
+		b = append(b, authVersion)
```
fortuna, line 148: Please keep the number here. It's harder to follow the protocol otherwise.
```suggestion
		b = append(b, 1)
```

transport/socks5/stream_dialer.go
```diff
@@ -55,80 +121,151 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
-	// VER = 5, CMD = 1 (connect), RSV = 0
-	b = append(b, 5, 1, 0)
-	// Destination address Address (ATYP, DST.ADDR, DST.PORT)
+	// VER = 5, CMD = 1 (connect), RSV = 0, DST.ADDR, DST.PORT
+	// +----+-----+-------+------+----------+----------+
+	// |VER | CMD |  RSV  | ATYP | DST.ADDR | DST.PORT |
+	// +----+-----+-------+------+----------+----------+
+	// | 1  |  1  | X'00' |  1   | Variable |    2     |
+	// +----+-----+-------+------+----------+----------+
+	b = append(b, socksProtocolVer, connectCommand, rsv)
```
fortuna, line 162: Please revert to have the numbers here.

transport/socks5/stream_dialer.go
```diff
@@ -23,23 +23,89 @@ import (
 	"github.com/Jigsaw-Code/outline-sdk/transport"
 )
 
+// bufferSize: The maximum buffer size is
+// 3 (1 socks version + 1 method selection + 1 methods)
+// + 1 (auth version) + 1 (username length) + 255 (username) + 1 (password length) + 255 (password)
+const (
```
fortuna, line 29: These constants are not helping readability. In fact they are hurting because it makes me go back an forth between the list and the code. Let's keep them inline like they were before. It's easier to follow the protocol that way.

transport/socks5/stream_dialer.go
```diff
@@ -23,23 +23,89 @@ import (
...
+	maxCredentialLength = 255
+	minCredentialLength = 1
+	socksProtocolVer    = 0x05
+	noAuthMethod        = 0x00
```
fortuna, line 33: It is, however, ok to declare protocol enumerations, like the auth method, in the socks5.go file.
But they should be named with the same prefix, like authMethodNoAuth and authMethodUserPass. Keep them private, since there's no need to expose them.

transport/socks5/stream_dialer.go
fortuna, line 140:
```suggestion
		b = append(buffer[:0], 5, 1, authMethodUserPass)
```

transport/socks5/stream_dialer.go
```diff
@@ -55,80 +121,151 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
-	// Buffer large enough for method and connect requests with a domain name address.
-	header := [3 + 4 + 256 + 2]byte{}
+	if c.credentials == nil {
+		// Method selection part: VER = 5, NMETHODS = 1, METHODS = 0 (no auth)
+		// +----+----------+----------+
+		// |VER | NMETHODS | METHODS  |
+		// +----+----------+----------+
+		// | 1  |    1     | 1 to 255 |
+		// +----+----------+----------+
+		b = append(buffer[:0], socksProtocolVer, numberOfAuthMethods, noAuthMethod)
```
fortuna, line 136:
```suggestion
		b = append(buffer[:0], 5, 1, authMethodNoAuth)
```

transport/socks5/stream_dialer.go
```diff
@@ -55,80 +121,151 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+	// 1. Read method response (VER, METHOD).
+	// +----+--------+
+	// |VER | METHOD |
+	// +----+--------+
+	// | 1  |   1    |
+	// +----+--------+
+	// buffer[0]: VER, buffer[1]: METHOD
+	// Reuse buffer for better performance.
+	if _, err = io.ReadFull(proxyConn, buffer[:2]); err != nil {
 		return nil, fmt.Errorf("failed to read method server response")
```
fortuna, line 187:
```suggestion
		return nil, fmt.Errorf("failed to read method server response: %w", err)
```

transport/socks5/stream_dialer.go
```diff
@@ -55,80 +121,151 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
 	}
-	if header[0] != 5 {
-		return nil, fmt.Errorf("invalid protocol version %v. Expected 5", header[0])
+	if buffer[0] != socksProtocolVer {
```
fortuna, line 189:
```suggestion
	if buffer[0] != 5 {
```

transport/socks5/stream_dialer.go
```diff
@@ -55,80 +121,151 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+	// |VER | REP |  RSV  | ATYP | BND.ADDR | BND.PORT |
+	// +----+-----+-------+------+----------+----------+
+	// | 1  |  1  | X'00' |  1   | Variable |    2     |
+	// +----+-----+-------+------+----------+----------+
+	// buffer[0]: VER
+	// buffer[1]: REP
+	// buffer[2]: RSV
+	// buffer[3]: ATYP
+	if _, err = io.ReadFull(proxyConn, buffer[:4]); err != nil {
+		fmt.Printf("failed to read connect server response: %v", err)
```
fortuna, line 231: Remove the print
```suggestion
```

transport/socks5/stream_dialer.go
```diff
@@ -55,80 +121,151 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
 	}
 
-	toRead := 0
-	switch header[3] {
+	// 4. Read address and length
+	var bndAddrLen int
+	switch buffer[3] {
 	case addrTypeIPv4:
-		toRead = 4
+		bndAddrLen = addrLengthIPv4
```
fortuna, line 247:
```suggestion
		bndAddrLen = 4
```

transport/socks5/stream_dialer.go
```diff
@@ -55,80 +121,151 @@ func (c *streamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
 	case addrTypeIPv6:
-		toRead = 16
+		bndAddrLen = addrLengthIPv6
```
fortuna, line 249:
```suggestion
		bndAddrLen = 16
```

transport/socks5/stream_dialer_test.go
```diff
@@ -164,3 +166,67 @@ func testExchange(tb testing.TB, listener *net.TCPListener, destAddr string, req
...
 	running.Wait()
 }
+
+func TestConnectWithoutAuth(t *testing.T) {
+	// Create a SOCKS5 server
+	server := socks5.NewServer()
+
+	// Create SOCKS5 proxy on localhost port 8000
+	go func() {
+		err := server.ListenAndServe("tcp", "127.0.0.1:8000")
```
fortuna, line 176: Use port zero to let the system pick the port for you to avoid conflicts. 
You will need to get the port somehow. Usually the listener will give you that. Perhaps the server gives you that too. Otherwise you will need to create the listener first, then pass to the server.

fortuna (changes requested):

transport/socks5/socks5.go
```diff
@@ -37,6 +37,11 @@ const (
 	ErrAddressTypeNotSupported       = ReplyCode(0x08)
 )
 
+const (
```
fortuna, line 40:
```suggestion
// SOCKS5 authentication methods, as specified in https://datatracker.ietf.org/doc/html/rfc1928#section-3
const (
```

transport/socks5/stream_dialer_test.go
```diff
@@ -164,3 +167,84 @@ func testExchange(tb testing.TB, listener *net.TCPListener, destAddr string, req
...
 	running.Wait()
 }
+
+func TestConnectWithoutAuth(t *testing.T) {
+	// Create a SOCKS5 server
+	server := socks5.NewServer()
+	host := "127.0.0.1"
+
+	// Create SOCKS5 proxy on localhost with a random port
+	listener, err := net.Listen("tcp", net.JoinHostPort(host, "0"))
```
fortuna, line 177: This is more readable:
```suggestion
	listener, err := net.Listen("tcp", "127.0.0.1:0")
```

transport/socks5/stream_dialer_test.go
```diff
@@ -164,3 +167,84 @@ func testExchange(tb testing.TB, listener *net.TCPListener, destAddr string, req
...
+		defer listener.Close()
+		fmt.Println("server is listening in the goroutin")
+		require.NoError(t, err)
+	}()
+
+	// wait for server to start
+	time.Sleep(10 * time.Millisecond)
+
+	// Create a SOCKS5 client
+	dialer, err := NewStreamDialer(&transport.TCPEndpoint{Address: net.JoinHostPort(host, strconv.Itoa(port))})
```
fortuna, line 194:
```suggestion
	dialer, err := NewStreamDialer(&transport.TCPEndpoint{Address: listener.Addr().String())
```

transport/socks5/stream_dialer_test.go
```diff
@@ -164,3 +167,84 @@ func testExchange(tb testing.TB, listener *net.TCPListener, destAddr string, req
...
+	require.NotNil(t, dialer)
+	require.NoError(t, err)
+
+	_, err = dialer.DialStream(context.Background(), net.JoinHostPort(host, strconv.Itoa(port)))
```
fortuna, line 198:
```suggestion
	_, err = dialer.DialStream(context.Background(), listener.Addr().String())
```

transport/socks5/stream_dialer_test.go
```diff
@@ -164,3 +167,84 @@ func testExchange(tb testing.TB, listener *net.TCPListener, destAddr string, req
...
+	server := socks5.NewServer()
+	host := "127.0.0.1"
+
+	// Create SOCKS5 proxy on localhost with a random port
+	listener, err := net.Listen("tcp", net.JoinHostPort(host, "0"))
+	require.NoError(t, err)
+
+	// Get the selected port number
+	port := listener.Addr().(*net.TCPAddr).Port
+
```
fortuna, line 182:
```suggestion
```

transport/socks5/stream_dialer_test.go
```diff
@@ -164,3 +167,84 @@ func testExchange(tb testing.TB, listener *net.TCPListener, destAddr string, req
...
+	go func() {
+		err := server.Serve(listener)
+		defer listener.Close()
+		fmt.Println("server is listening in the goroutin")
```
fortuna, line 186: Remove Prints. It pollutes the output. You can use [TB.Log](https://pkg.go.dev/testing#TB) though.

transport/socks5/stream_dialer_test.go
```diff
@@ -164,3 +167,84 @@ func testExchange(tb testing.TB, listener *net.TCPListener, destAddr string, req
...
+			"testusername": "testpassword",
+		},
+	}
+	server := socks5.NewServer(
+		socks5.WithAuthMethods([]socks5.Authenticator{cator}),
+	)
+
+	// Create SOCKS5 proxy on localhost with a random port
+	host := "127.0.0.1"
+	listener, err := net.Listen("tcp", net.JoinHostPort(host, "0"))
```
fortuna, line 215:
```suggestion
	listener, err := net.Listen("tcp", "127.0.0.1:0")
```

transport/socks5/stream_dialer.go
```diff
@@ -23,26 +23,59 @@ import (
...
+
+func NewCredentials(username, password string) (Credentials, error) {
+	var c Credentials
+	usernameBytes := []byte(username)
+	passwordBytes := []byte(password)
+
+	if len(usernameBytes) > 255 {
+		return c, errors.New("username exceeds 255 bytes")
+	}
+	if len(usernameBytes) == 0 {
```
fortuna, line 41: Let's remove this check. I don't think this is an RFC requirement. Same for the password.

fortuna (changes requested):

transport/socks5/stream_dialer.go
```diff
@@ -23,26 +23,54 @@ import (
...
 }
 
-var _ transport.StreamDialer = (*streamDialer)(nil)
+var _ transport.StreamDialer = (*StreamDialer)(nil)
+
+func (c *StreamDialer) SetCredentials(username, password []byte) error {
+	if len(username) > 255 {
+		return errors.New("username exceeds 255 bytes")
+	}
+	if len(username) == 0 {
```
fortuna, line 53: It's ok for the fields to be empty. Please remove the zero checks.

fortuna (commented):
Just one minor thing

transport/socks5/stream_dialer.go
```diff
@@ -23,26 +23,54 @@ import (
 	"github.com/Jigsaw-Code/outline-sdk/transport"
 )
 
+// https://datatracker.ietf.org/doc/html/rfc1929
+// Credentials can be nil, and that means no authentication.
+type Credentials struct {
```
fortuna, line 28: Make it private

fortuna (approved):

fortuna (approved):

PR #193, opened 2024-03-05
[EXPERIMENTAL] prototyping a next-generation cross-platform outline client with the sdk

daniellacosse:
currently have a go web app rendering on web, android:
image

and apple (mac catalyst):
image

next steps: 
- [x] add htmx to the go web app
- [x] integrate the sdk into it
- [x] add a field where we can paste in an access key
- [ ] stub out vpn platform service
- [ ] write platform service on each box

### DIFF
--- a/x/examples/outline-vpn-app/app/start.go
+++ b/x/examples/outline-vpn-app/app/start.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"log"
 	"net/http"
@@ -11,7 +12,6 @@ import (
 	"github.com/Jigsaw-Code/outline-sdk/x/config"
 	"github.com/Jigsaw-Code/outline-sdk/x/mobileproxy"
 	"github.com/a-h/templ"
-	"github.com/labstack/gommon/bytes"
 )
 
 var appAddress = "[::1]:8080"
@@ -19,7 +19,7 @@ var proxyAddress = "[::1]:8181"
 var systemAddress = "[::1]:8282"
 
 var proxyDisconnectTimeoutSecond = 5
-var systemTunnelEndpoint = "/vpn/"
+var systemTunnelEndpoint = fmt.Sprintf("%s/vpn/", systemAddress)
 
 type VPNController struct {
 	ID     string
--- a/x/examples/outline-vpn-app/app/server_view.templ
+++ b/x/examples/outline-vpn-app/app/server_view.templ
@@ -35,19 +35,41 @@ templ serverView() {
           background-color: hsl(170, 3%, 95%);
           flex-grow: 1;
           display: flex;
+          flex-direction: column;
           align-items: center;
           justify-content: center;
         }
 
+        section {
+          width: 100%;
+          max-width: 540px;
+          padding: 1rem;
+        }
+
+        textarea {
+          background-color: hsl(170, 8%, 85%);
+          padding: 1rem;
+          width: 100%;
+          border: 2px solid hsl(170, 5%, 45%);
+          border-radius: 0.5rem;
+          min-height: 350px;
+          font-family: monospace;
+          word-wrap: break-word;
+        }
+
+        textarea:focus {
+          outline: none;
+          border-color: hsl(170, 40%, 55%);
+        }
+
         button {
           cursor: pointer;
-          border: 0.5rem dashed hsl(170, 5%, 45%);
           background-color: hsl(170, 8%, 85%);
-          width: clamp(240px, 33vw, 480px);
-          height: clamp(240px, 33vw, 480px);
-          border-radius: 100%;
+          border-radius: 0.5rem;
           color: hsl(170, 5%, 45%);
           text-align: center;
+          width: 100%;
+          height: 5rem;
           font-size: 1.5rem;
           font-weight: bold;
           transition-duration: 1000ms;
@@ -86,6 +108,10 @@ templ serverView() {
       <main>
         <header><h1>outline</h1></header>
         <article>
+          <section>
+            <textarea placeholder="paste your access key here">
+            </textarea>
+          </section>
           <section>
             @connectionButton()
           </section>


x/examples/outline-vpn-app/fullstack_app/start.go
```diff
@@ -0,0 +1,124 @@
...
+	http.ListenAndServe(APP_ADDRESS, nil)
+}
+
+type VPNController struct {
+	proxy  *mobileproxy.Proxy
+	tunnel io.Reader
+}
+
+func (vpn *VPNController) handleConnectionEndpoint(responseWriter http.ResponseWriter, request *http.Request) {
+	endpoint, encryptionKey := parseStaticShadowsocksAccessKey(request.URL.Path)
```
jyyi1, line 41: Check for any existing connections first? But still, there might be race conditions here, we can deal with that later.

```suggestion
	if vpn.proxy != nil {
		http.Error(responseWriter, "already connected", http.StatusConflict)
		return
	}

	endpoint, encryptionKey := parseStaticShadowsocksAccessKey(request.URL.Path)
```

x/examples/outline-vpn-app/fullstack_app/start.go
```diff
@@ -0,0 +1,124 @@
...
+
+	if err != nil {
+		http.Error(responseWriter, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	vpn.proxy = proxy
+
+	// TODO: why this code isn't very good at all!
+	tunnel, err := os.Open("tunnel.sock")
```
jyyi1, line 60: Check `err` ?

x/examples/outline-vpn-app/fullstack_app/start.go
```diff
@@ -0,0 +1,124 @@
...
+	vpn.tunnel = tunnel
+
+	// TODO: implement system vpn tunnel service
```
jyyi1, line 63: Do you mean implement the service using "unix domain socket" (`Open("tunnel.sock")`)? This is only supported by Linux based OS (Ubuntu/Android), maybe not even Android.

x/examples/outline-vpn-app/fullstack_app/start.go
```diff
@@ -0,0 +1,124 @@
...
+	http.NewRequest("POST", SYSTEM_TUNNEL_ENDPOINT, vpn.tunnel)
+
+	writeTemplate(disconnectionButton, responseWriter)
+}
+
+func (vpn *VPNController) handleDisconnectionEndpoint(responseWriter http.ResponseWriter, _ *http.Request) {
+	http.NewRequest("DELETE", SYSTEM_TUNNEL_ENDPOINT, vpn.tunnel)
+
+	vpn.proxy.Stop(PROXY_DISCONNECT_TIMEOUT)
+
```
jyyi1, line 80: Set the proxy to `nil`? So we will be able to check the connection state in the next "connection"? As it will also mark the object for garbage collection

```suggestion
	vpn.proxy = nil
	vpn.tunnel.Close()
```

x/examples/outline-vpn-app/fullstack_app/start.go
```diff
@@ -0,0 +1,124 @@
...
+	writeTemplate(connectionButton, responseWriter)
+}
+
+func parseStaticShadowsocksAccessKey(staticAccessKey string) (transport.StreamEndpoint, *shadowsocks.EncryptionKey) {
```
jyyi1, line 84: It seems this function can be replaced by [`config.NewStreamDialer(staticAccessKey)`](https://pkg.go.dev/github.com/Jigsaw-Code/outline-sdk/x@v0.0.0-20240223000159-142376ee10ea/config#NewStreamDialer)? We just need to make sure the key starts with `ss://` and doesn't contain any pipe characters (`|`). And then returns the `StreamDialer` to the caller.

x/examples/outline-vpn-app/fullstack_app/start.go
jyyi1: We probably need to have an abstraction, like `VPNRoutingConfigurer`. And provide different implementations on different platforms.

x/examples/outline-vpn-app/app/start.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/Jigsaw-Code/outline-sdk/x/mobileproxy"
+	"github.com/a-h/templ"
+)
+
+var appAddress = ":8080"
+var proxyAddress = ":8181"
+var systemAddress = ":8282"
+
+var proxyDisconnectTimeoutSecond = 5
+var systemTunnelEndpoint = "${systemAddress}/tunnel/${proxyAddress}"
```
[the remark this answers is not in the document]
jyyi1: `fmt.Sprintf("%s/tunnel/%s", systemAddress, proxyAddress)`: https://pkg.go.dev/fmt#Sprintf

x/examples/outline-vpn-app/app/start.go
```diff
@@ -0,0 +1,87 @@
...
+	"io"
+	"net/http"
+	"os"
+
+	"github.com/Jigsaw-Code/outline-sdk/x/config"
+	"github.com/Jigsaw-Code/outline-sdk/x/mobileproxy"
+	"github.com/a-h/templ"
+)
+
+var appAddress = ":8080"
```
jyyi1, line 14: I guess `appPort` would be more accurate?

x/examples/outline-vpn-app/app/start.go
```diff
@@ -0,0 +1,87 @@
...
+	tunnel, err := os.Open("tunnel.sock")
+
+	if err != nil {
+		// do something
+	}
+
+	vpn.tunnel = tunnel
+
+	// TODO: implement system vpn tunnel service
+	// => POST /tunnel/URL forward all non-local traffic to that URL
```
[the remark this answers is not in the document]
jyyi1: The Tunnel here is more like a system VPN configuration, typically it involves the source and the target. The source can be simple, like "all non-local TCP and UDP traffic from the system", or it can be complicated, such as "all UDP traffic destined to `8.8.8.8` from a specific app". The target would typically be represented by `IP:port` instead of a `URL`.

For the simplicity, I think we can start with `source=all TCP & UDP` and `target=<outline-server-ip>:<port>`. And we need to provide different implementations for different OS.

x/examples/outline-vpn-app/app/start.go
```diff
@@ -0,0 +1,87 @@
...
+	// => DELETE /tunnel/URL gracefully shuts down the tunnel
+
+	// => apple box uses VPN API
+	// => kotlin box uses SSH tunnel (for now)
+
+	// ! these boxes will be reusable across VPN apps !
+	http.NewRequest("POST", systemTunnelEndpoint, vpn.tunnel)
```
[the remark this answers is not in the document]
jyyi1: Looks good, but please adjust the API according to the comment above.

The request might return permission error, in this case the app should handle it (e.g., navigate to the permission approval settings page in Android, or pop-up a dialog in Linux to enter the root password).

x/examples/outline-vpn-app/app/start.go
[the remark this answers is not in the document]
jyyi1: Yes. At least for Windows and Linux there are no differences. But on Android and iOS, we need to adapt to the VPN API provided by the system.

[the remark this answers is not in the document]
jyyi1: That's right.

PR #203, opened 2024-03-23
example(outline-cli): configure system DNS when `resolv.conf` does not exist

zeslava:
It's ok if no original resolv.conf file found


jyyi1 (approved):
Thanks for the contribution!

jyyi1 (changes requested):
Thanks for the contribution.

But this logic might not work when there is `resolv.head.conf`, but not `resolv.conf`. Previously, only `resolv.conf.outlinecli.backup` will be created and only `resolv.conf` will be changed; but now both `resolv.head.conf` and `resolv.conf` will be changed.

We might need to add a special logic to `setSystemDNSServer` to handle the situation that neither `resolv.head.conf` nor `resolv.conf` exist in the system. In that case, we just need to create a new `resolv.conf` with Outline's setting, and when we `restoreSystemDNSServer`, we need to delete the newly created `resolv.conf` (because no backups are available).

x/examples/outline-cli/dns_linux.go
```diff
@@ -27,30 +28,58 @@ const (
...
+	err := os.WriteFile(original, data, 0o644)
+	if err != nil {
+		return func() {}, fmt.Errorf("failed to write DNS config file '%s': %w", original, err)
 	}
-	return nil
+
+	return restore, err
 }
 
 func restoreFileIfExists(backup, original string) {
```
jyyi1, line 85: And finally, you may simply loop though all entries in `backups` here (we don't need to take any `backup` or `original` parameters in this function any more):

```
for _, backup := range backups {
	if _, err := os.Stat(backup.backup); err != nil {
		// if err is notexist
		//   simplify remove backup.original
		logging.Warn.Printf("filed to get info of DNS config backup file '%s': %v\n", backup, err)
		return
	}
	if err := os.Rename(backup.backup, backup.original); err != nil {
		logging.Err.Printf("failed to restore DNS config from backup '%s' to '%s': %v\n", backup.backup, backup.original, err)
		return
	}
	logging.Info.Printf("DNS config restored from '%s' to '%s'\n", backup.backup, backup.original)
}
```

jyyi1 (changes requested):
Thanks a lot for the update! I believe we can even simplify the logic by introducing a global `backups` variable here. So we don't need to have any restoration callback functions, which is a little bit harder to understand.

zeslava:
@jyyi1 fixed

x/examples/outline-cli/dns_linux.go
```diff
@@ -27,40 +28,77 @@ const (
 	resolvConfHeadBackupFile = "/etc/resolv.head.outlinecli.backup"
 )
 
+// systemDNSBackup info about systemdns backup
```
jyyi1, line 31: This comment doesn't seem to be necessary.

```suggestion
```

x/examples/outline-cli/dns_linux.go
```diff
@@ -27,40 +28,77 @@ const (
...
 func setSystemDNSServer(serverHost string) error {
 	setting := []byte(`# Outline CLI DNS Setting
 # The original file has been renamed as resolv[.head].outlinecli.backup
 nameserver ` + serverHost + "\n")
 
-	if err := backupAndWriteFile(resolvConfFile, resolvConfBackupFile, setting); err != nil {
-		return err
+	err := backupAndWriteFile(resolvConfFile, resolvConfBackupFile, setting)
+	if err != nil {
+		return fmt.Errorf("dns backup: %w", err)
```
jyyi1, line 45: I'd rather simply return the error, because we have provided enough error details in `backupAndWriteFile` function.

```suggestion
		return err
```

x/examples/outline-cli/dns_linux.go
```diff
@@ -27,40 +28,77 @@ const (
...
 	}
-	return backupAndWriteFile(resolvConfHeadFile, resolvConfHeadBackupFile, setting)
-}
 
-func restoreSystemDNSServer() {
-	restoreFileIfExists(resolvConfBackupFile, resolvConfFile)
-	restoreFileIfExists(resolvConfHeadBackupFile, resolvConfHeadFile)
+	err = backupAndWriteFile(resolvConfHeadFile, resolvConfHeadBackupFile, setting)
+	if err != nil {
+		return fmt.Errorf("dns head backup: %w", err)
```
jyyi1, line 50: Similar to above.

```suggestion
		return err
```

jyyi1: Also if we are simply returning the `err` here, we can just do `return backupAndWriteFile(...)` directly.

x/examples/outline-cli/dns_linux.go
```diff
@@ -27,40 +28,77 @@ const (
...
 func backupAndWriteFile(original, backup string, data []byte) error {
-	if err := os.Rename(original, backup); err != nil {
-		return fmt.Errorf("failed to backup DNS config file '%s' to '%s': %w", original, backup, err)
+	if _, err := os.Stat(original); err == nil {
+		// original file exist - move it into backup
+		if err := os.Rename(original, backup); err != nil {
+			return fmt.Errorf("failed to backup DNS config file '%s' to '%s': %w", original, backup, err)
+		}
+	} else if !errors.Is(err, os.ErrNotExist) { // if not exist - it's ok, just write to it
+		return errors.New("check original file exists")
```
jyyi1, line 63:
```suggestion
		return fmt.Errorf("failed to check the existence of DNS config file '%s': %w", original, err)
```

x/examples/outline-cli/dns_linux.go
```diff
@@ -27,40 +28,77 @@ const (
...
+	// save data to original
+	if err := os.WriteFile(original, data, 0o644); err != nil {
 		return fmt.Errorf("failed to write DNS config file '%s': %w", original, err)
 	}
+
+	systemDNSBackups = append(systemDNSBackups, systemDNSBackup{
+		original: original,
+		backup:   backup,
+	})
+
```
jyyi1, line 75: I'd suggest to put these lines before `os.WriteFile`. Because we have already backed up the original DNS config, even though we failed to write to the new file, we still need to copy the backup file back to original when we try to reset the DNS config.

x/examples/outline-cli/dns_linux.go
```diff
@@ -27,40 +28,77 @@ const (
...
+			logging.Info.Printf("DNS config restored from '%s' to '%s'\n", backup.backup, backup.original)
+		} else if errors.Is(err, os.ErrNotExist) {
+			// backup not exist - just remove original, because it's created by ourselves
+			if err := os.Remove(backup.original); err != nil {
+				logging.Err.Printf("removing original DNS config file '%s': %v", backup.original, err)
+				continue
+			}
+			logging.Info.Printf("DNS config removed '%s'\n", backup.original)
+		} else {
+			logging.Err.Printf("checking backup '%s': %v", backup.backup, err)
```
jyyi1, line 101:
```suggestion
			logging.Err.Printf("checking backup '%s': %v\n", backup.backup, err)
```

x/examples/outline-cli/dns_linux.go
```diff
@@ -27,40 +28,77 @@ const (
...
+					backup.original,
+					err,
+				)
+				continue
+			}
+			logging.Info.Printf("DNS config restored from '%s' to '%s'\n", backup.backup, backup.original)
+		} else if errors.Is(err, os.ErrNotExist) {
+			// backup not exist - just remove original, because it's created by ourselves
+			if err := os.Remove(backup.original); err != nil {
+				logging.Err.Printf("removing original DNS config file '%s': %v", backup.original, err)
```
jyyi1, line 96:
```suggestion
				logging.Err.Printf("failed to remove Outline DNS config file '%s': %v\n", backup.original, err)
```

x/examples/outline-cli/dns_linux.go
```diff
@@ -27,40 +28,77 @@ const (
...
+				continue
+			}
+			logging.Info.Printf("DNS config removed '%s'\n", backup.original)
```
jyyi1, line 99:
```suggestion
			logging.Info.Printf("Outline DNS config '%s' has been removed\n", backup.original)
```

x/examples/outline-cli/ipv6_linux.go
```diff
@@ -40,7 +40,7 @@ func enableIPv6(enabled bool) (bool, error) {
 	} else {
 		disabledStr[0] = '1'
 	}
-	if err := os.WriteFile(disableIPv6ProcFile, disabledStr, 0644); err != nil {
+	if err := os.WriteFile(disableIPv6ProcFile, disabledStr, 0o644); err != nil {
```
jyyi1, line 43: 👍! Thanks.

jyyi1: We can keep this change. Though it's not related to DNS, but it's CLI.

x/examples/outline-cli/dns_linux.go
zeslava: done

jyyi1 (approved):
Thanks @zeslava , the code looks good to me!

jyyi1:
Hi @zeslava , on last thing, could you fetch the latest `main` and merge it to this PR? It will unblock the failing CI so I can merge this PR. Thanks.

PR #242, opened 2024-05-28
fix(test): wrong url used in expired certificate test

jyyi1:
The `TestRevoked` should be `TestExpired` because the reason `x509.Expired` in `CertificateInvalidError` stands for an expired certificate.

Revoked certificate test is tricky, because it relies on the OS implementation (e.g., [macOS](https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/crypto/x509/root_darwin.go;l=71), [Windows](https://cs.opensource.google/go/go/+/master:src/crypto/x509/root_windows.go;l=203)), the error is weakly typed and different OS will report different error messages. For now the revocation test case for Linux and Windows are skipped because the certificate revocation list is not up-to-date on these two platforms.

The failing CI is blocking all PRs including #203, #226, and #241 .

### DIFF
--- a/transport/tls/stream_dialer_test.go
+++ b/transport/tls/stream_dialer_test.go
@@ -17,6 +17,7 @@ package tls
 import (
 	"context"
 	"crypto/x509"
+	"runtime"
 	"testing"
 
 	"github.com/Jigsaw-Code/outline-sdk/transport"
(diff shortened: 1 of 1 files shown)


transport/tls/stream_dialer_test.go
```diff
@@ -44,10 +44,10 @@ func TestUntrustedRoot(t *testing.T) {
 	require.ErrorAs(t, err, &certErr)
 }
 
-func TestRevoked(t *testing.T) {
+func TestExpired(t *testing.T) {
 	sd, err := NewStreamDialer(&transport.TCPDialer{})
 	require.NoError(t, err)
-	_, err = sd.DialStream(context.Background(), "revoked.badssl.com:443")
```
fortuna, line 50: I explicitly wanted to test for revoked here. I don't think the issue is with the original test. It seems like `revoked.badssl.com` no longer serves an expired cert, so we don't return an error. Chrome is considering it safe.
image

Can you instead comment it out or use `t.Skip("Skipping until revoked.badssl.com is fixed")` in addition to adding the expired test?

fortuna (changes requested):

PR #256, opened 2024-07-01
WIP: Adding UDP support to Socks5

amircybersec:
Currently Socks5 in the SDK does not support UDP. The goal of this PR is to add UDP support which would be useful for getting measurements on UDP blocking.


amircybersec:
duplicate of https://github.com/Jigsaw-Code/outline-sdk/pull/257

PR #257, opened 2024-07-02
Adding UDP support to Socks5 dialer

amircybersec:
This PR add packet dialing (UDP) support to Socks5 dialer. The test runs a local UDP echo (ping-pong) server and socks5 server and routes the UDP packet through the local proxy server to the echo server and back. This was tested with several remote servers including soax service.


amircybersec:
@fortuna I'd appreciate it if you could review this PR.

transport/socks5/packet_dialer.go
```diff
@@ -0,0 +1,166 @@
...
+func (c *packetConn) SetReadDeadline(t time.Time) error {
+	return c.pc.SetReadDeadline(t)
+}
+
+func (c *packetConn) SetWriteDeadline(t time.Time) error {
+	return c.pc.SetWriteDeadline(t)
+}
+
+func (c *packetConn) Read(b []byte) (int, error) {
+	// TODO: read header
```
fortuna, line 57:
```suggestion
```

transport/socks5/packet_dialer.go
```diff
@@ -0,0 +1,166 @@
...
+	buffer := make([]byte, 65536) // Maximum size for UDP packet
+	n, err := c.pc.Read(buffer)
+	if err != nil {
+		return 0, err
+	}
+	// Minimum size of header is 10 bytes
+	if n < 10 {
+		return 0, fmt.Errorf("invalid SOCKS5 UDP packet: too short")
+	}
```
fortuna, line 66: It make me nervous that you are not immediately creating a slice with the proper boundaries. Let's work on the pkt slice instead of buffer, since buffer is bigger than the packet.
```suggestion
	}
	pkt := buffer[:n]
```

transport/socks5/packet_dialer.go
```diff
@@ -0,0 +1,166 @@
...
+		return 0, fmt.Errorf("fragmentation is not supported")
+	}
+
+	atyp := buffer[3]
+	addrLen := 0
+	switch atyp {
+	case addrTypeIPv4:
+		addrLen = net.IPv4len
+	case addrTypeIPv6:
+		addrLen = net.IPv6len
```
fortuna, line 85: This is a SOCKS5 definition. `net.IPv4len` is a implementation definition of the net package that just happens to match.

Let's use the numbers from the RFC to avoid confusion.

```suggestion
	case addrTypeIPv4:
		addrLen = 4
	case addrTypeIPv6:
		addrLen = 16
```

transport/socks5/packet_dialer.go
```diff
@@ -0,0 +1,166 @@
...
+	if rsv[0] != 0x00 || rsv[1] != 0x00 {
+		return 0, fmt.Errorf("invalid reserved bytes: expected 0x0000, got %#x%#x", rsv[0], rsv[1])
+	}
+
+	frag := buffer[2]
+	if frag != 0 {
+		return 0, fmt.Errorf("fragmentation is not supported")
+	}
+
+	atyp := buffer[3]
```
fortuna, line 79: Please extract the address reading into its own method that is shared with the stream code.
It should take a reader, like `readAddress(reader) (string, error)`.

That way it will work for both functions.

We can also add tests for the address reading.

transport/socks5/stream_dialer.go
```diff
@@ -194,44 +199,68 @@ func (c *StreamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
 	if buffer[1] != 0 {
-		return nil, ReplyCode(buffer[1])
+		return nil, "", ReplyCode(buffer[1])
 	}
 
-	// 4. Read address and length
+	// 4. Read BND.ADDR.
+	host := ""
 	var bndAddrLen int
 	switch buffer[3] {
```
fortuna, line 217: Let's introduce a readAddress as I mention in the packet code.

transport/socks5/packet_dialer.go
```diff
@@ -0,0 +1,166 @@
...
+func (c *packetConn) Read(b []byte) (int, error) {
+	// TODO: read header
+	buffer := make([]byte, 65536) // Maximum size for UDP packet
+	n, err := c.pc.Read(buffer)
+	if err != nil {
+		return 0, err
+	}
+	// Minimum size of header is 10 bytes
+	if n < 10 {
+		return 0, fmt.Errorf("invalid SOCKS5 UDP packet: too short")
```
fortuna, line 65:
```suggestion
		return 0, errors.New("invalid SOCKS5 UDP packet: too short")
```

transport/socks5/packet_dialer.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// Start parsing the header
+	rsv := buffer[:2]
+	if rsv[0] != 0x00 || rsv[1] != 0x00 {
+		return 0, fmt.Errorf("invalid reserved bytes: expected 0x0000, got %#x%#x", rsv[0], rsv[1])
+	}
+
+	frag := buffer[2]
+	if frag != 0 {
+		return 0, fmt.Errorf("fragmentation is not supported")
```
fortuna, line 76:
```suggestion
		return 0, errors.New("fragmentation is not supported")
```

transport/socks5/packet_dialer.go
```diff
@@ -0,0 +1,166 @@
...
+	// Calculate the start position of the actual data
+	headerLength := 4 + addrLen + 2 // RSV (2) + FRAG (1) + ATYP (1) + ADDR (variable) + PORT (2)
+	if n < headerLength {
+		return 0, fmt.Errorf("invalid SOCKS5 UDP packet: header too short")
+	}
+
+	// Copy the payload into the provided buffer
+	payloadLength := n - headerLength
+	if payloadLength > len(b) {
+		// maybe raise an error to indicate that the provided buffer is too small?
```
fortuna, line 102: Let's return `io.ErrShortBuffer`

transport/socks5/packet_dialer.go
```diff
@@ -0,0 +1,166 @@
...
+	return c.pc.SetReadDeadline(t)
+}
+
+func (c *packetConn) SetWriteDeadline(t time.Time) error {
+	return c.pc.SetWriteDeadline(t)
+}
+
+func (c *packetConn) Read(b []byte) (int, error) {
+	// TODO: read header
+	buffer := make([]byte, 65536) // Maximum size for UDP packet
```
fortuna, line 58: This is an unnecessary allocation.
For the payload, we can read directly from the conn to the output buffer at the end.

Just use the connection directly, reading the fields as needed, like we do in the stream code.

transport/socks5/packet_dialer.go
```diff
@@ -0,0 +1,166 @@
...
+	n, err := c.pc.Read(buffer)
```
fortuna, line 59: Let's not read the entire packet. Instead, read each field as you go, like we do in the stream code.

transport/socks5/packet_dialer.go
```diff
@@ -0,0 +1,166 @@
...
+		payloadLength = len(b)
+	}
+	copy(b, buffer[headerLength:n])
+
+	return payloadLength, nil
+}
+
+func (c *packetConn) Write(b []byte) (int, error) {
+	// Encapsulate the payload in a SOCKS5 UDP packet
+	header := []byte{
```
fortuna, line 112: Allocate it so it can hold at least an IPv4 address without requiring reallocation.

fortuna (changes requested):

transport/socks5/packet_dialer.go
amircybersec: @fortuna Thanks for the feedback. I looked further into the suggested approach but it appears that I can only do one single Read and subsequent read operations time out. 

In UDP (User Datagram Protocol), each call to Read or ReadFrom on a UDP socket reads a single datagram from the network stack. This is fundamentally how UDP operates at the socket level; each datagram (packet) sent by a sender corresponds to one atomic message that should be read in its entirety by the receiver. If I attempt to read part of the message, I lose the rest of the message. This behaviour is different from TCP connection where I can read incrementally from the Connection. Since the package size is not known beforehand, I have to allocate maximum buffer size to fix the entire datagram.

transport/socks5/packet_dialer.go
amircybersec: @fortuna got it! will change that accordingly. Please however note that this will be anyways reallocated for IPv6 and domain name since the exact size is not know beforehand.

transport/socks5/packet_dialer.go
amircybersec: @fortuna there's some intricacy here. I like the idea of introducing `readAddress` that we can use both in stream and packet code; however the `Read` behavior is different between TCP and UDP. In TCP, I can read from buffer in chunks and increments but in UDP I can only perform one Read operation per datagram(packet) -- Please see related comment below. We can potentially pass reference to a buffer to `readAddress` (instead of reader) but in Stream code we are not reading all data at once into the buffer.

transport/socks5/packet_dialer.go
amircybersec: @fortuna some relevant discussion: https://forum.golangbridge.org/t/reading-from-udp-socket-blocks-even-if-there-is-data/3290/2

transport/socks5/packet_dialer.go
amircybersec: @fortuna The allocation is necessary for UDP since we need to read the whole datagram at once otherwise we lose the rest of the data. please see related comments below.

fortuna: You are right. Let's use the slicepool like we do with shadowsocks.

transport/socks5/packet_dialer.go
fortuna: I meant, we can wrap a buffer with Reader with `bytes.NewReader(buffer)`

fortuna (changes requested):
Almost there.

fortuna (changes requested):

transport/socks5/stream_dialer.go
```diff
@@ -30,23 +34,25 @@ type credentials struct {
 	password []byte
 }
 
-// NewStreamDialer creates a [transport.StreamDialer] that routes connections to a SOCKS5
+// NewDialer creates a [transport.StreamDialer] that routes connections to a SOCKS5
 // proxy listening at the given [transport.StreamEndpoint].
-func NewStreamDialer(endpoint transport.StreamEndpoint) (*StreamDialer, error) {
-	if endpoint == nil {
+func NewDialer(streamEndpoint transport.StreamEndpoint) (*Dialer, error) {
```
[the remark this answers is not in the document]
fortuna: That's because it seems SOCKS5 provides a ListenPacket interface, not a DialPacket interface. This is the same as Shadowsocks.
So the code should align with that, and we can create a dialer from a listener, just like we do with Shaodwsocks.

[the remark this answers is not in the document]
amircybersec: @fortuna I have been digging into this; Yes, I believe that is the way to go too. Regarding the address included in the association request, it seems like that it must be either "0.0.0.0:0" for the unbound case or the address the client is listening on for the bound case (not the remote destination address). I can confirm this via the Socks server code lines below:

https://github.com/things-go/go-socks5/blob/d67aeab3543eddb5ea51a821ba6879e89f981d3e/handle.go#L222

Also, in the following test a listener is setup for the client and the client listening address it included in the associate request:

https://github.com/things-go/go-socks5/blob/d67aeab3543eddb5ea51a821ba6879e89f981d3e/server_test.go#L173

I am going to define `NewPacketListener, and `PacketListen` similar to Shadowsocks.

transport/socks5/stream_dialer.go
amircybersec: I added `readAddress` and pushed the latest changes.

transport/socks5/packet_dialer.go
amircybersec: @fortuna the latest change uses `slicepool` for better use of memory.

transport/socks5/packet_dialer.go
amircybersec: @fortuna got it. thanks for clarification. I implemented readAddress accordingly.

transport/socks5/packet_dialer.go
```diff
@@ -0,0 +1,216 @@
...
+	default:
+		return 0, fmt.Errorf("unknown address type %#x", atyp)
+	}
+
+	pkt = pkt[4:] // Skip the header
+	addr := pkt[:addrLen]
+
+	pkt = pkt[addrLen:] // Skip the address
+	port := binary.BigEndian.Uint16(pkt[:2])
+	fmt.Printf("Received packet from %d:%d\n", addr, port)
```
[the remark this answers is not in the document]
amircybersec: @fortuna sure! I removed most of logging prints but will do a clean up at the end.

transport/socks5/packet_dialer.go
```diff
@@ -0,0 +1,216 @@
...
+	return errors.Join(c.sc.Close(), c.pc.Close())
+}
+
+// DialPacket creates a packet [net.Conn] via SOCKS5.
+func (d *Dialer) DialPacket(ctx context.Context, dstAddr string) (net.Conn, error) {
+	netDstAddr, err := transport.MakeNetAddr("udp", dstAddr)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse address: %w", err)
+	}
+	sc, bindAddr, err := d.request(ctx, CmdUDPAssociate, "0.0.0.0:0")
```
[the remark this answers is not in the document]
amircybersec: @fortuna please refer to the comment below with pointers to server code and tests.

transport/socks5/stream_dialer.go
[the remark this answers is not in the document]
amircybersec: @fortuna also looking at the `packet.go` and implementation of `boundPacketConn` in the SDK, It seems like the bound definition in the SDK is based on remote address binding where as in SOCK5 spec, binding refers to local address (client address) binding, i.e. the server only accepts packets from bound origin IP/Port arriving at the proxy side listening socket. I guess the bound connection concept can be applicable to both server and client side sockets and is a rather broad concept. 

```go
// Read implements [net.Conn].Read.
func (c *boundPacketConn) Read(packet []byte) (int, error) {
	for {
		n, remoteAddr, err := c.PacketConn.ReadFrom(packet)
		if err != nil {
			return n, err
		}
		if remoteAddr.String() != c.remoteAddr.String() {
			continue
		}
		return n, nil
	}
}
```

transport/socks5/packet_dialer.go
```diff
@@ -0,0 +1,218 @@
...
+	fullPacket := append(header, b...)
+	return c.pc.Write(fullPacket)
+}
+
+func (c *packetConn) Close() error {
+	return errors.Join(c.sc.Close(), c.pc.Close())
+}
+
+// DialPacket creates a packet [net.Conn] via SOCKS5.
+func (d *Dialer) DialPacket(ctx context.Context, dstAddr string) (net.Conn, error) {
```
fortuna, line 140: This should be renamed to `ListenPacket`, not take a destination address and return a `net.PacketConn`.

fortuna (changes requested):

transport/socks5/packet_dialer.go
```diff
@@ -0,0 +1,218 @@
...
+
+func (c *packetConn) SetReadDeadline(t time.Time) error {
+	return c.pc.SetReadDeadline(t)
+}
+
+func (c *packetConn) SetWriteDeadline(t time.Time) error {
+	return c.pc.SetWriteDeadline(t)
+}
+
+func (c *packetConn) Read(b []byte) (int, error) {
```
fortuna, line 66: This needs to be replaced with the UDP ReadFrom, which returns the packet address. See https://github.com/Jigsaw-Code/outline-sdk/blob/e157b30b8d094e008d1e38d40f58967274717c1c/transport/shadowsocks/packet_listener.go#L91

transport/socks5/packet_dialer.go
```diff
@@ -0,0 +1,218 @@
...
+	payloadLength := n - headerLength
+	if payloadLength > len(b) {
+		return 0, io.ErrShortBuffer
+	}
+	copy(b, buffer[headerLength:n])
+
+	return payloadLength, nil
+}
+
+func (c *packetConn) Write(b []byte) (int, error) {
```
fortuna, line 115: This needs to be UDP WriteTo instead, taking a destination address. See https://github.com/Jigsaw-Code/outline-sdk/blob/e157b30b8d094e008d1e38d40f58967274717c1c/transport/shadowsocks/packet_listener.go#L69

fortuna (changes requested):

transport/socks5/packet_dialer.go
amircybersec: @fortuna I implemented and tested this change.

transport/socks5/packet_dialer.go
amircybersec: @fortuna change applied.

transport/socks5/packet_dialer.go
amircybersec: @fortuna done :)

transport/socks5/socks5.go
```diff
@@ -119,3 +127,47 @@ func appendSOCKS5Address(b []byte, address string) ([]byte, error) {
 	b = binary.BigEndian.AppendUint16(b, uint16(portNum))
 	return b, nil
 }
+
+func readAddress(reader io.Reader) (string, int, error) {
```
fortuna, line 131: Please add tests for this function.

fortuna, line 131: There's no need to return the bytes read. Remove.

transport/socks5/socks5.go
```diff
@@ -119,3 +127,47 @@ func appendSOCKS5Address(b []byte, address string) ([]byte, error) {
...
+	var buffer [1 + 1 + 256]byte
+	addrLen := 0
+	_, err := io.ReadFull(reader, buffer[:1])
+	if err != nil {
+		return "", 0, fmt.Errorf("failed to read address type: %w", err)
+	}
+	// Read the address type
+	switch buffer[0] {
+	case addrTypeIPv4:
+		addrLen = 4
```
fortuna, line 144: Put the logic to read each type in their respective case.

transport/socks5/socks5.go
```diff
@@ -119,3 +127,47 @@ func appendSOCKS5Address(b []byte, address string) ([]byte, error) {
...
+		addrLen = int(buffer[1])
+	default:
+		return "", 0, fmt.Errorf("unknown address type %#x", buffer[0])
+	}
+	// Read host address
+	_, err = io.ReadFull(reader, buffer[:addrLen])
+	if err != nil {
+		return "", 0, fmt.Errorf("failed to read address: %w", err)
+	}
+	host := net.IP(buffer[:addrLen]).String()
```
fortuna, line 163: This logic doesn't work with the domain name type. We need tests.

transport/socks5/packet_listener.go
```diff
@@ -0,0 +1,182 @@
...
+
+// DialPacket creates a packet [net.Conn] via SOCKS5.
+func (c *Client) ListenPacket(ctx context.Context) (net.PacketConn, error) {
+	// Connect to the SOCKS5 server and perform UDP association
+	sc, bindAddr, err := c.request(ctx, CmdUDPAssociate, "0.0.0.0:0")
+	if err != nil {
+		return nil, err
+	}
+
+	host, port, err := net.SplitHostPort(bindAddr)
```
fortuna, line 149: We are converting the host to a string, then joining with the port, just to undo all of that here.

Let's make request and readAddress return the address and port separately, and not convert the IP to string.

I suggest creating structs like:

```go
type AddressHost struct {
  // Only one of these will be filled
  ip *netip.Addr
  domain string
}

type Address struct {
  host AddressHost
  port uint16
}
```

transport/socks5/stream_dialer.go
```diff
@@ -30,23 +30,26 @@ type credentials struct {
...
-	return &StreamDialer{proxyEndpoint: endpoint, cred: nil}, nil
+	return &Client{se: streamEndpoint, cred: nil}, nil
 }
 
-type StreamDialer struct {
-	proxyEndpoint transport.StreamEndpoint
-	cred          *credentials
+type Client struct {
+	se   transport.StreamEndpoint
+	pe   transport.PacketEndpoint
```
fortuna, line 44: Remove the PacketEndpoint, it's not needed.

transport/socks5/packet_listener.go
```diff
@@ -0,0 +1,182 @@
...
+
+	proxyConn, err := c.pe.ConnectPacket(ctx)
+	if err != nil {
+		sc.Close()
+		return nil, fmt.Errorf("could not connect to packet endpoint: %w", err)
+	}
+	return &packetConn{pc: proxyConn, sc: sc, dstAddr: proxyConn.RemoteAddr()}, nil
+}
+
+func (c *Client) EnablePacketListener(endpoint transport.PacketEndpoint) error {
```
fortuna, line 176: Delete this function. It's not needed.

transport/socks5/packet_listener.go
```diff
@@ -0,0 +1,182 @@
...
+			return nil, fmt.Errorf("failed to parse tcp address: %w", err)
+		}
+		host = schost
+	}
+
+	packetEndpoint := &transport.PacketDialerEndpoint{Dialer: c.pd, Address: net.JoinHostPort(host, port)}
+	err = c.EnablePacketListener(packetEndpoint)
+	if err != nil {
+		return nil, fmt.Errorf("failed to enable packet listener: %w", err)
+	}
```
fortuna, line 166: Remove this.

transport/socks5/packet_listener.go
```diff
@@ -0,0 +1,182 @@
...
+
+	proxyConn, err := c.pe.ConnectPacket(ctx)
```
fortuna, line 168:
```suggestion
	proxyConn, err := packetEndpoint.ConnectPacket(ctx)
```

transport/socks5/stream_dialer.go
```diff
@@ -65,15 +68,14 @@ func (c *StreamDialer) SetCredentials(username, password []byte) error {
...
-// the connect requests in one packet, to avoid an additional roundtrip.
-// The returned [error] will be of type [ReplyCode] if the server sends a SOCKS error reply code, which
-// you can check against the error constants in this package using [errors.Is].
-func (c *StreamDialer) DialStream(ctx context.Context, remoteAddr string) (transport.StreamConn, error) {
-	proxyConn, err := c.proxyEndpoint.ConnectStream(ctx)
+func (c *Client) EnablePacket(packetDialer transport.PacketDialer) {
+	c.pd = packetDialer
+}
+
+func (c *Client) request(ctx context.Context, cmd byte, dstAddr string) (transport.StreamConn, string, error) {
```
fortuna, line 75: This function is doing a bit too much. Can you make it take an io.ReadWriter and move the connection to the caller instead? This way we simplify the return value and separate the concerns.

transport/socks5/packet_listener.go
```diff
@@ -0,0 +1,182 @@
...
+		return 0, nil, fmt.Errorf("failed to resolve address: %w", err)
+	}
+	return payloadLength, addr, nil
+}
+
+func (p *packetConn) WriteTo(b []byte, addr net.Addr) (int, error) {
+	// Encapsulate the payload in a SOCKS5 UDP packet as specified in
+	// https://datatracker.ietf.org/doc/html/rfc1928#section-7
+	// The minimum preallocated header size (10 bytes)
+	header := make([]byte, 10)
```
fortuna, line 121: You should allocate space for at least IPv4 + the input buffer size.

transport/socks5/packet_listener.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	// Do something with address?
+	address, addrLen, err := readAddress(bytes.NewReader(pkt[3:]))
+	if err != nil {
+		return 0, nil, fmt.Errorf("failed to read address: %w", err)
+	}
+
+	// Calculate the start position of the actual data
+	headerLength := 4 + addrLen + 2 // RSV (2) + FRAG (1) + ATYP (1) + ADDR (variable) + PORT (2)
```
fortuna, line 98: Don't do this calculation. Use the bytes.Buffer instead. It will keep track of the bytes for you.
There should be no direct read on the buffer after you create the Buffer.

Later you can call `copy` with `Buffer.Bytes()`

transport/socks5/packet_listener.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	// Copy the payload into the provided buffer
+	payloadLength := n - headerLength
+	if payloadLength > len(b) {
+		return 0, nil, io.ErrShortBuffer
+	}
+	copy(b, buffer[headerLength:n])
+
+	addr, err = net.ResolveUDPAddr("udp", address)
```
fortuna, line 110: Don't resolver the address. Use https://pkg.go.dev/github.com/Jigsaw-Code/outline-sdk/transport#MakeNetAddr instead.

transport/socks5/packet_listener.go
fortuna: Actually, let's use the buffer pool here too. Or else we will be doing allocations per write, which is too much.

fortuna (changes requested):

x/config/socks5.go
```diff
@@ -27,18 +27,46 @@ func wrapStreamDialerWithSOCKS5(innerSD func() (transport.StreamDialer, error),
...
 		}
 	}
-	return dialer, nil
+
+	return client, nil
+}
+
+func wrapPacketDialerWithSOCKS5(_ func() (transport.StreamDialer, error), innerPD func() (transport.PacketDialer, error), configURL *url.URL) (transport.PacketDialer, error) {
+	// Use a TCP dialer to connect to the SOCKS5 server.
+	sd := &transport.TCPDialer{}
```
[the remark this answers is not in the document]
amircybersec: @fortuna my main concern with using `innerSD` is that since this only used for UDP association request and if the base dialer is something other than plain `&transport.TCPDialer{}` (for example if base dialer is shadowsocks) it can cause problems.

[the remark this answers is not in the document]
fortuna: Why would it cause problems? Without using the innerSD you are not properly composing the transports. You can't do multi-hop without that.

transport/socks5/packet_dialer.go
amircybersec: @fortuna change applied.

x/config/socks5.go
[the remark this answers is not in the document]
amircybersec: @fortuna nevermind. You are right since we are not transferring data using innerSD and we just use it to dial to socks5 server and send connect/associate command. I applied the change.

transport/socks5/socks5.go
amircybersec: @fortuna readAddress changed to returned address struct and tests added.

amircybersec: @fortuna removed.

transport/socks5/socks5.go
amircybersec: @fortuna updated readAddress accordingly.

transport/socks5/socks5.go
amircybersec: @fortuna updated code and added tests

transport/socks5/packet_listener.go
amircybersec: @fortuna introduced address data type accordingly.

transport/socks5/stream_dialer.go
amircybersec: @amircybersec thanks for catching this. yes it is redundant.

transport/socks5/packet_listener.go
amircybersec: done.

transport/socks5/packet_listener.go
amircybersec: done.

transport/socks5/packet_listener.go
amircybersec: done.

transport/socks5/stream_dialer.go
amircybersec: @fortuna good point. I refactored the code accordingly.

transport/socks5/packet_listener.go
amircybersec: @fortuna makes sense. now using buffer pool.

transport/socks5/packet_listener.go
amircybersec: @fortuna done!

transport/socks5/packet_listener.go
amircybersec: @fortuna change applied.

transport/socks5/socks5_test.go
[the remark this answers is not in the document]
amircybersec: I added the benchmark for `readAddress`. I tried to lower the number of allocations to 3 but it increased the `B/op` by a lot. current `B/op` seems reasonable though. 
```sh
➜  socks5 git:(fortuna-socks) ✗ go test -bench=. -benchmem
IP IP goos: darwin
goarch: arm64
pkg: github.com/Jigsaw-Code/outline-sdk/transport/socks5
BenchmarkReadAddr/IPv4-8                12668048                93.64 ns/op      56 B/op          4 allocs/op
BenchmarkReadAddr/IPv6-8                11770556               101.1 ns/op      68 B/op           4 allocs/op
BenchmarkReadAddr/Domain-8               9883328               120.9 ns/op      80 B/op           5 allocs/op
PASS
ok      github.com/Jigsaw-Code/outline-sdk/transport/socks5     4.254s
```

transport/socks5/socks5_test.go
```diff
@@ -15,12 +15,99 @@
...
+	tests := []struct {
+		name    string
+		input   []byte
+		want    *address
+		wantErr bool
+	}{
+
+		{
+			name:    "IPv4 Example",
+			input:   append([]byte{addrTypeIPv4}, append(net.IPv4(192, 168, 1, 1).To4(), []byte{0x01, 0xF4}...)...),
```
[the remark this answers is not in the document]
amircybersec: @fortuna inputs are updated and simplified as much as possible.

transport/socks5/socks5.go
```diff
@@ -79,6 +87,29 @@ const (
...
+	Name string // fully-qualified domain name
+	IP   net.IP
+	Port int
+}
+
+func (a *address) Network() string { return "socks5" }
+
+// Address returns a string suitable to dial; prefer returning IP-based
+// address, fallback to Name
+func (a *address) String() string {
```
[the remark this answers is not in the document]
amircybersec: I changed this to standalone method. We use this in a few places to convert address type to string.

transport/socks5/socks5.go
```diff
@@ -79,6 +87,29 @@ const (
...
 
+// address is a SOCKS-specific address.
+// Either Name or IP is used exclusively.
+type address struct {
+	Name string // fully-qualified domain name
+	IP   net.IP
+	Port int
+}
+
+func (a *address) Network() string { return "socks5" }
```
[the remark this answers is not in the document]
amircybersec: done.

transport/socks5/socks5.go
```diff
@@ -79,6 +87,29 @@ const (
 	addrTypeIPv6 = 0x04
 )
 
+// address is a SOCKS-specific address.
+// Either Name or IP is used exclusively.
+type address struct {
+	Name string // fully-qualified domain name
+	IP   net.IP
+	Port int
```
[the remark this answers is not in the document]
amircybersec: done.

transport/socks5/packet_listener.go
```diff
@@ -0,0 +1,184 @@
...
+		bindAddr.IP = net.ParseIP(schost)
+	}
+
+	packetEndpoint := &transport.PacketDialerEndpoint{Dialer: c.pd, Address: bindAddr.String()}
+	proxyConn, err := packetEndpoint.ConnectPacket(ctx)
+	if err != nil {
+		sc.Close()
+		return nil, fmt.Errorf("could not connect to packet endpoint: %w", err)
+	}
+	return &packetConn{pc: proxyConn, sc: sc, dstAddr: proxyConn.RemoteAddr()}, nil
```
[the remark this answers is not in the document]
amircybersec: done.

transport/socks5/packet_listener.go
```diff
@@ -0,0 +1,184 @@
...
+	sc      transport.StreamConn
+}
+
+var _ net.PacketConn = (*packetConn)(nil)
+
+func (p *packetConn) LocalAddr() net.Addr {
+	return p.pc.LocalAddr()
+}
+
+func (p *packetConn) RemoteAddr() net.Addr {
```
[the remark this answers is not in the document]
amircybersec: done.

transport/socks5/packet_listener.go
```diff
@@ -0,0 +1,184 @@
...
+	if ipAddr := bindAddr.IP; ipAddr != nil && ipAddr.IsUnspecified() {
+		schost, _, err := net.SplitHostPort(sc.RemoteAddr().String())
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse tcp address: %w", err)
+		}
+		bindAddr.IP = net.ParseIP(schost)
+	}
+
+	packetEndpoint := &transport.PacketDialerEndpoint{Dialer: c.pd, Address: bindAddr.String()}
+	proxyConn, err := packetEndpoint.ConnectPacket(ctx)
```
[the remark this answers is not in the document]
amircybersec: done.

transport/socks5/packet_listener.go
```diff
@@ -0,0 +1,184 @@
...
+// clientUDPBufferSize is the maximum supported UDP packet size in bytes.
+const clientUDPBufferSize = 16 * 1024
+
+// udpPool stores the byte slices used for storing packets.
+var udpPool = slicepool.MakePool(clientUDPBufferSize)
+
+type packetConn struct {
+	dstAddr net.Addr
+	pc      net.Conn
+	sc      transport.StreamConn
```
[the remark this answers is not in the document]
amircybersec: done.

transport/socks5/packet_dialer.go
```diff
@@ -0,0 +1,215 @@
...
+		domainAddrLen := make([]byte, 1)
+		_, err := io.ReadFull(reader, domainAddrLen)
+		if err != nil {
+			return "", 0, fmt.Errorf("failed to read domain address length: %w", err)
+		}
+		addrLen = int(domainAddrLen[0])
+	default:
+		return "", 0, fmt.Errorf("unknown address type %#x", addrType[0])
+	}
+	host := make([]byte, addrLen)
```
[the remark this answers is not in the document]
fortuna: Please address this security concern. Just change the type of addrLen to byte. That will impose a 255 size limit.

transport/socks5/packet_listener.go
```diff
@@ -0,0 +1,183 @@
...
+	buffer := lazySlice.Acquire()
+	defer lazySlice.Release()
+
+	n, err := p.pc.Read(buffer)
+	if err != nil {
+		return 0, nil, err
+	}
+	// Minimum packet size
+	if n < 10 {
+		return 0, nil, fmt.Errorf("invalid SOCKS5 UDP packet: too short")
```
fortuna, line 73:
```suggestion
		return 0, nil, errors.New("invalid SOCKS5 UDP packet: too short")
```

transport/socks5/packet_listener_test.go
```diff
@@ -0,0 +1,107 @@
...
+	echoServer := setupUDPEchoServer(t, echoServerAddr)
+	defer echoServer.Close()
+
+	// Create a socks server to proxy "ping" message
+	cator := socks5.UserPassAuthenticator{Credentials: socks5.StaticCredentials{
+		"testusername": "testpassword",
+	}}
+	proxySrv := socks5.NewServer(
+		socks5.WithAuthMethods([]socks5.Authenticator{cator}),
+		//socks5.WithLogger(socks5.NewLogger(log.New(os.Stdout, "socks5: ", log.LstdFlags))),
```
fortuna, line 32: Remove comment.

transport/socks5/packet_listener_test.go
```diff
@@ -0,0 +1,107 @@
...
+	)
+	// Start listening
+	proxyServerAddress := "127.0.0.1:12355"
+	go func() {
+		err := proxySrv.ListenAndServe("tcp", proxyServerAddress)
+		require.NoError(t, err)
+	}()
+	time.Sleep(10 * time.Millisecond)
```
fortuna, line 40: What is this sleep for? We can probably avoid it. It's usually bad to use sleep in tests, since variations in execution time can break the test.

transport/socks5/packet_listener_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// Connect, auth and connec to local server
```
fortuna, line 42: It's misspelled, and the local server is the proxy?


Perhaps:
```suggestion
	// Connect to local proxy, auth and start the PacketConn.
```

transport/socks5/packet_listener_test.go
```diff
@@ -0,0 +1,107 @@
...
+	// Create a socks server to proxy "ping" message
+	cator := socks5.UserPassAuthenticator{Credentials: socks5.StaticCredentials{
+		"testusername": "testpassword",
+	}}
+	proxySrv := socks5.NewServer(
+		socks5.WithAuthMethods([]socks5.Authenticator{cator}),
+		//socks5.WithLogger(socks5.NewLogger(log.New(os.Stdout, "socks5: ", log.LstdFlags))),
+	)
+	// Start listening
+	proxyServerAddress := "127.0.0.1:12355"
```
fortuna, line 35: No fixed ports in tests please!

Use the listener local address instead to get the port assigned by the OS.

transport/socks5/packet_listener_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/things-go/go-socks5"
+)
+
+func TestSOCKS5Associate(t *testing.T) {
+	// Create a local listener
+	// This creates a UDP server that responded to "ping"
+	// message with "pong" response
+	locIP := net.ParseIP("127.0.0.1")
+	// Create a local listener
+	echoServerAddr := &net.UDPAddr{IP: locIP, Port: 12199}
```
fortuna, line 22: No hard-coded ports for local servers. Retrieve it from the listener local address instead.

transport/socks5/packet_listener_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// Send "ping" message
+	_, err = conn.WriteTo([]byte("ping"), echoServerAddr)
+	require.NoError(t, err)
+	// max wait time for response
+	conn.SetDeadline(time.Now().Add(time.Second))
+	response := make([]byte, 1024)
+	n, addr, err := conn.ReadFrom(response)
+	require.Equal(t, echoServerAddr, addr)
+	//conn.SetDeadline(time.Time{})
```
fortuna, line 61: ?

transport/socks5/packet_listener_test.go
```diff
@@ -0,0 +1,107 @@
...
+	require.NotNil(t, client)
+	require.NoError(t, err)
+	err = client.SetCredentials([]byte("testusername"), []byte("testpassword"))
+	require.NoError(t, err)
+	client.EnablePacket(&transport.UDPDialer{})
+	conn, err := client.ListenPacket(context.Background())
+	require.NoError(t, err)
+	defer conn.Close()
+
+	// Send "ping" message
```
fortuna, line 53: Punctuation
```suggestion
	// Send "ping" message.
```

transport/socks5/packet_listener_test.go
```diff
@@ -0,0 +1,107 @@
...
+	_, err = conn.WriteTo([]byte("ping"), echoServerAddr)
+	require.NoError(t, err)
+	// max wait time for response
```
fortuna, line 56: Proper capitalization.
```suggestion
	// Max wait time for response.
```

transport/socks5/packet_listener_test.go
```diff
@@ -0,0 +1,107 @@
...
+	response := make([]byte, 1024)
+	n, addr, err := conn.ReadFrom(response)
+	require.Equal(t, echoServerAddr, addr)
+	//conn.SetDeadline(time.Time{})
+	require.NoError(t, err)
+	require.Equal(t, []byte("pong"), response[:n])
+}
+
+func TestUDPLoopBack(t *testing.T) {
+	// Create a local listener
```
fortuna, line 67:
```suggestion
	// Create a local listener.
```

transport/socks5/packet_listener_test.go
```diff
@@ -0,0 +1,107 @@
...
+	locIP := net.ParseIP("127.0.0.1")
+	// Create a local listener
```
fortuna, line 69:
```suggestion
	// Create a local listener.
```

transport/socks5/packet_listener_test.go
```diff
@@ -0,0 +1,107 @@
...
+	echoServerAddr := &net.UDPAddr{IP: locIP, Port: 12199}
```
fortuna, line 70: No hard-coded server port

transport/socks5/packet_listener_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+func setupUDPEchoServer(t *testing.T, serverAddr *net.UDPAddr) *net.UDPConn {
+	server, err := net.ListenUDP("udp", serverAddr)
+	require.NoError(t, err)
+	go func() {
+		buf := make([]byte, 2048)
+		for {
+			n, remote, err := server.ReadFrom(buf)
+			if err != nil {
+				//log.Printf("Error reading: %v", err)
```
fortuna, line 92: remove

transport/socks5/packet_listener_test.go
```diff
@@ -0,0 +1,107 @@
...
+				return
+			}
+			//log.Printf("Received %s from %s\n", buf[:n], remote.String())
```
fortuna, line 95: Remove

transport/socks5/socks5.go
```diff
@@ -119,3 +149,45 @@ func appendSOCKS5Address(b []byte, address string) ([]byte, error) {
...
+		if _, err := io.ReadFull(r, addr[:]); err != nil {
+			return nil, err
+		}
+		address.IP = netip.AddrFrom16(addr)
+	case addrTypeDomainName:
+		if _, err := r.Read(addrType[:]); err != nil {
+			return nil, err
+		}
+		addrLen := int(addrType[0])
+		fqdn := make([]byte, addrLen)
```
fortuna, line 179: Make sure addrLen is of type byte, to limit the allocation.
```suggestion
		addrLen := addrType[0]
		fqdn := make([]byte, int(addrLen))
```

transport/socks5/socks5_test.go
```diff
@@ -15,12 +15,160 @@
...
+			name:    "IPv4 Example",
+			input:   []byte{addrTypeIPv4, 192, 168, 1, 1, 0x01, 0xF4},
+			want:    &address{IP: netip.AddrFrom4([4]byte{192, 168, 1, 1}), Port: 500},
+			wantErr: false,
+		},
+		{
+			name: "IPv6 Full",
+			input: []byte{
+				addrTypeIPv6,
+				0x20, 0x01, 0x0d, 0xb8, // first 4 bytes of the IPv6 address
```
fortuna, line 45: Why don't you use decimal for IPv4?

transport/socks5/socks5_test.go
```diff
@@ -15,12 +15,160 @@
...
+				0x0b,                                                  // Length of the domain name "example.com" which is 11 characters
+				'e', 'x', 'a', 'm', 'p', 'l', 'e', '.', 'c', 'o', 'm', // The domain name "example.com"
+				0x23, 0x28, // Port number 9000 in hexadecimal
+			},
+			want:    &address{Name: "example.com", Port: 9000},
+			wantErr: false,
+		},
+		{
+			name:    "Domain Long",
+			input:   append([]byte{addrTypeDomainName, 0x3B}, append([]byte("very-long-domain-name-used-for-testing-purposes.example.com"), []byte{0x00, 0x50}...)...),
```
fortuna, line 93:
```suggestion
			input:   append([]byte{addrTypeDomainName, 0x3B}, append([]byte("very-long-domain-name-used-for-testing-purposes.example.com"), 0x00, 0x50)...),
```

transport/socks5/stream_dialer.go
```diff
@@ -30,23 +30,25 @@ type credentials struct {
 	password []byte
 }
 
-// NewStreamDialer creates a [transport.StreamDialer] that routes connections to a SOCKS5
+// NewDialer creates a [transport.StreamDialer] that routes connections to a SOCKS5
```
fortuna, line 33:
```suggestion
// NewClient creates a SOCKS5 client that routes connections to a SOCKS5
```

transport/socks5/stream_dialer.go
```diff
@@ -65,23 +67,13 @@ func (c *StreamDialer) SetCredentials(username, password []byte) error {
...
-	if err != nil {
-		return nil, fmt.Errorf("could not connect to SOCKS5 proxy: %w", err)
-	}
-	dialSuccess := false
-	defer func() {
-		if !dialSuccess {
-			proxyConn.Close()
-		}
-	}()
+func (c *Client) EnablePacket(packetDialer transport.PacketDialer) {
```
fortuna, line 70:
```suggestion
// EnablePacket enables the use of the [Client] as a [transport.PacketListener]. It takes the [transport.PacketDialer] used to connect to the SOCKS5 packet endpoint.
func (c *Client) EnablePacket(packetDialer transport.PacketDialer) {
```

transport/socks5/stream_dialer.go
```diff
@@ -206,32 +198,47 @@ func (c *StreamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+func (c *Client) connectAndRequest(ctx context.Context, cmd byte, dstAddr string) (transport.StreamConn, *address, error) {
+	proxyConn, err := c.se.ConnectStream(ctx)
+	if err != nil {
+		return nil, nil, fmt.Errorf("could not connect to SOCKS5 proxy: %w", err)
+	}
+
+	// Close the connection if an error occurs in any subsequent steps
+	// to avoid keeping the connection open.
+	defer func(conn transport.StreamConn) {
+		if conn != nil && err != nil {
```
fortuna, line 220: conn is never nil, because you are passing it as a parameter on successful ConnectStream.

I think it's cleaner if you just call close in the case of request error below, given the logic is simple.

transport/socks5/stream_dialer.go
```diff
@@ -206,32 +198,47 @@ func (c *StreamDialer) DialStream(ctx context.Context, remoteAddr string) (trans
...
+			conn.Close()
+		}
+	}(proxyConn)
+
+	bindAddr, err := c.request(proxyConn, cmd, dstAddr)
+	if err != nil {
+		return nil, nil, err
+	}
```
fortuna, line 228:
```suggestion
	bindAddr, err := c.request(proxyConn, cmd, dstAddr)
	if err != nil {
		proxyConn.Close()
		return nil, nil, err
	}
```

fortuna (changes requested):

transport/socks5/socks5.go
amircybersec: @fortuna we need to make sure `addrLen` is under 255 to prevent large memory allocation. The following code prevents this and also throws a meaningful error message is length is more than 255
 
```go
		addrLen := int(addrType[0])
		if addrLen > 255 {
			return nil, fmt.Errorf("domain name length = %v is over 255", addrLen)
		}
		fqdn := make([]byte, addrLen)
		if _, err := io.ReadFull(r, fqdn); err != nil {
			return nil, err
		}
```

transport/socks5/packet_listener_test.go
amircybersec: @fortuna thanks for pointing this out. We had already done this in StreamDialer tests. I removed the hard coded ports in tests.

transport/socks5/packet_listener_test.go
amircybersec: removed.

transport/socks5/packet_listener_test.go
amircybersec: @fortuna This was there to ensure server is ready before continuing with tests. I agree if the server takes longer than 10 ms to start listening, the test is going to fail. I added an async method to wait until the server is ready to make this more robust. I also had this sleep in `stream_dialer_test.go` and replaced it with `waitUntilServerReady` there as well.

```go 
func waitUntilServerReady(addr string, ready chan<- bool) {
	for {
		conn, err := net.Dial("tcp", addr)
		if err == nil {
			conn.Close()
			ready <- true
			return
		}
		time.Sleep(10 * time.Millisecond)
	}
}
```

transport/socks5/socks5.go
fortuna: This is not needed if you keep the type of addrLen as byte.
Why are you converting to int?

transport/socks5/packet_listener_test.go
```diff
@@ -0,0 +1,123 @@
...
+	go func() {
+		err := proxySrv.Serve(listener)
+		defer listener.Close()
+		require.NoError(t, err)
+	}()
+
+	// Wait until the server is ready.
+	ready := make(chan bool, 1)
+	go waitUntilServerReady(proxyServerAddress, ready)
+	<-ready
```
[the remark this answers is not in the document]
amircybersec: @fortuna please see my above comments.

transport/socks5/packet_listener_test.go
```diff
@@ -0,0 +1,123 @@
...
+	for {
+		conn, err := net.Dial("tcp", addr)
+		if err == nil {
+			conn.Close()
+			ready <- true
+			return
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+}
```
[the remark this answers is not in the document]
amircybersec: @fortuna I ran the tests without the delay/sleep and `waitUntilServerReady`, and the test would fail around 10~20% of the time. Probably listen action when starting the server is async and may not have started even though the `ListenAndServe` return no error.  In the `things-go/go-socks5` server tests, there's this delay:

https://github.com/things-go/go-socks5/blob/d67aeab3543eddb5ea51a821ba6879e89f981d3e/server_test.go#L58

`waitUntilServerReady` is a better approach though that just using sleep since it guarantees a listening server.

transport/socks5/packet_listener_test.go
amircybersec: @fortuna I missed an important detail. Yes you are right. I forgot I changed the server call from `ListenAndServe` to `Serve` and I begin to listen separately with the last change. My test result was related to previous implementation that combined Listen and Serve into one call. I removed all the waits and re-tested the code. I ran it around 50 times and it never failed without the delay.

transport/socks5/packet_listener_test.go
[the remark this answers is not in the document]
amircybersec: @fortuna Neverming, you were right. My test result was related to change before removing hard-coded ports and passing a listener to the server. With this change, the wait is no longer necessary. I re-tested and never observed an error.

amircybersec:
@fortuna thanks again for the feedback. I applied the latest rounds of comments.

transport/socks5/packet_listener_test.go
```diff
@@ -0,0 +1,123 @@
...
+	)
+
+	// Create SOCKS5 proxy on localhost with a random port
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err)
+	proxyServerAddress := listener.Addr().String()
+
+	go func() {
+		err := proxySrv.Serve(listener)
+		defer listener.Close()
```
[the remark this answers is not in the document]
fortuna: This still needs to be moved.

fortuna (commented):
Looks good. Just one minor thing. Your test is leaking a goroutine

transport/socks5/packet_listener_test.go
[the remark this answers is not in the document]
amircybersec: @fortuna There's a small detail here. I can move `defer listener.Close()` to right after where we start listening but when the test finishes and listener closes the server goroutine throws a closed connection error. Unfortunately the server does not accept a context parameter to cancel it; Once solution is to consider an exception for server's `net.ErrClosed` error and basically ignore it. Not the cleanest approach though. Do you have any thoughts on this?

```go
	require.NoError(t, err)
	defer listener.Close()
	proxyServerAddress := listener.Addr().String()

	go func() {
		err := proxySrv.Serve(listener)
		if !errors.Is(err, net.ErrClosed) && err != nil {
			require.NoError(t, err) // Assert no error if it's not the expected close error
		}
	}()
```

fortuna (approved):
Thanks for all the changes! This is great work!

amircybersec:
@fortuna I resolved the merge conflict in `x/go.mod` and `x/go.sum`; I am requiring following version in mod file as before but the build is still failing. It seems like `x/config/socks5` is still pointing older `transport/socks5` package.

`github.com/Jigsaw-Code/outline-sdk v0.0.17-0.20240708052545-4f51b366fe75`

Not sure what I am doing wrong...

PR #268, opened 2024-08-25
add ssconf support

pmoieni:
this PR adds support for `ssconf` transport URL to Outline CLI. I was not sure what should be done with the `prefix` field so just left it untouched. I was not able to test the implementation either.

> OutlineVPN CLI (experimental)
[INFO] 2024/08/25 15:12:11 updated global IPv6 support: false
[INFO] 2024/08/25 15:12:11 updated global IPv6 support: true
[ERROR] 2024/08/25 15:12:11 failed to create OutlineDevice: failed to create TCP dialer: failed to decode host string [ss:///*redacted*/@/*redacted*/:8080]: illegal base64 data at input byte 2

using the official access keys I get the above error.
no such issue while using `ssconf` key:
> OutlineVPN CLI (experimental)
[INFO] 2024/08/25 15:58:38 updated global IPv6 support: false
[INFO] 2024/08/25 15:58:41 remote server cannot handle UDP traffic, switch to DNS truncate mode.
[INFO] 2024/08/25 15:58:41 routing traffic from 10.233.233.1 to 10.233.233.2/32 through nic 9
[INFO] 2024/08/25 15:58:41 routing traffic via gw 10.233.233.2 through nic 9...
[INFO] 2024/08/25 15:58:41 ip rule 'from all not to /*redacted*/ via table 233' created

but it doesn't seem to establish any kind of connection at all

OS: OpenSuse Tumbleweed

### DIFF
--- a/x/examples/outline-cli/outline_device.go
+++ b/x/examples/outline-cli/outline_device.go
@@ -46,11 +46,18 @@ type OutlineDevice struct {
 var configToDialer = config.NewDefaultConfigToDialer()
 
 func NewOutlineDevice(transportConfig string) (od *OutlineDevice, err error) {
-	transportConfig, err = formatConfig(transportConfig)
+	if err := validateConfig(transportConfig); err != nil {
+		return nil, err
+	}
+	parsed, err := url.Parse(transportConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse config URL: %w", err)
+	}
+	transportConfig, err = formatConfig(parsed)
 	if err != nil {
 		return nil, err
 	}
-	ip, err := resolveShadowsocksServerIPFromConfig(transportConfig)
+	ip, err := resolveShadowsocksServerIPFromHostname(parsed.Hostname())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


jyyi1:
Thanks so much for this PR! We're currently migrating dynamic key feature into Go in Outline Client as well. To ensure the best integration, we'd appreciate a bit more time to review your PR and explore potential code sharing opportunities.

pmoieni:
@jyyi1  looking forward to that, ready to improve the code if necessary.

alexander-potemkin:
> Hi @alexander-potemkin , thanks for being interested in the SDK. The good news is, `ssconf://` is pretty simple. You just need to call `http.Client.Get` on the URL (but change `ssconf` to `https`), and then parse the content or use it directly if it's a static key like `ss://...`.
> 
> The tricky part is supporting the new YAML format in Outline Client. The config formats have diverged: [Client](https://github.com/Jigsaw-Code/outline-apps/blob/master/client/go/outline/config/module.go), [SDK](https://github.com/Jigsaw-Code/outline-sdk/blob/main/x/configurl/module.go). So, we need to either copy the Outline Client's logic to SDK, or just support the legacy format in the SDK.
> 
> This means, you might not be able to share the same `ssconf://` key between Outline Client and the SDK for now. But hopefully, we can merge these two formats soon.

Hi @jyyi1 , thanks for the prompt response and detailed explanation in there!

I didn't look down the source code, nor I know the roadmap / the goals, but from the outside view it looks like it would be best of all if Client would be just a wrapper around SDK, so that everyone integrating SDK would benefit.

I hear you that it is not possible at the moment, is there some chances someone will take a look at that problem (and, probably, that PR) anytime soon?

Btw, does it also mean that WebSocket feature is not accessible via SDK as well? I believe - it is not, but thought I would better double check...

Legion112:
I see here support for `yaml` is missing.

PR #298, opened 2024-10-14
fix: try base64 legacy decoding for maximum compatibility

amircybersec:
Currently <b> Outline SDK </b> config parser fails for strings encoded in legacy base64 mode such as the example below:

```
ss://YWVzLTEyOC1nY206c2hhZG93c29ja3M=@210.100.100.100:443
```

Current Outline client is however able to parse and decode the above URL. For consistent behaviour between Outline SDK and Outline client, I have added a secondary decoding step if current method fails. 

Current aproach:
```go
encoding := base64.URLEncoding.WithPadding(base64.NoPadding).DecodeString(userInfo)
```

Backup approach:
```go
encoding := base64.StdEncoding.WithPadding(base64.NoPadding).DecodeString(userInfo)
```


sbruens (changes requested):
Do you mind adding a unit test for this case?

amircybersec:
@sbruens I added two tests; Please let me know if more tests are needed.

sbruens (approved):

PR #324, opened 2024-11-10
Allow for a connection to check if it's sending bytes

PeterZhizhin:
This will be helpful for some strategies:
- Split: check for bytes to get sent before sending next packet
- Disorder, Fake: wait for bytes to get sent before changing TTL back


x/sockopt/is_sending_bytes_linux.go
```diff
@@ -0,0 +1,42 @@
...
+		return false, err
+	}
+	var result bool
+	syscallConn.Control(func(fd uintptr) {
+		innerResult, innerErr := isSocketFdSendingBytes(int(fd))
+		result = innerResult
+		err = innerErr
+	})
+	return result, err
+}
```
derlaft, line 42:
```suggestion
func isConnectionSendingBytes(conn *net.TCPConn) (result bool, err error) {
	syscallConn, err := conn.SyscallConn()
	if err != nil {
		return false, err
	}
	syscallConn.Control(func(fd uintptr) {
		result, err = isSocketFdSendingBytes(int(fd))
	})
	return
}
```

x/sockopt/is_sending_bytes_linux.go
```diff
@@ -0,0 +1,42 @@
...
+	tcpInfo, err := unix.GetsockoptTCPInfo(int(fd), unix.IPPROTO_TCP, unix.TCP_INFO)
+	if err != nil {
+		return false, err
+	}
+
+	// 1 == TCP_ESTABLISHED, but for some reason not available in the package
+	if tcpInfo.State != 1 {
+		// If the connection is not established, the socket is not sending bytes
+		return false, nil
+	}
```
derlaft, line 21: could you confirm this does what we supposed the bye-dpi code does? I'm starting having some doubts looking [at the list of states](https://elixir.bootlin.com/linux/v6.10.6/source/include/net/tcp_states.h)

x/sockopt/is_sending_bytes_linux.go
```diff
@@ -0,0 +1,42 @@
...
+package sockopt
+
+import (
+	"net"
+
+	"golang.org/x/sys/unix"
+)
+
+func isSocketFdSendingBytes(fd int) (bool, error) {
+	tcpInfo, err := unix.GetsockoptTCPInfo(int(fd), unix.IPPROTO_TCP, unix.TCP_INFO)
```
derlaft, line 12: `int(int)`?

x/sockopt/is_sending_bytes_not_implemented.go
```diff
@@ -0,0 +1,17 @@
+//go:build !linux
```
derlaft, line 1: maybe it's worth noting that this is not even needed on some platforms (we know of mac/ios already)

x/sockopt/is_sending_bytes_linux.go
PeterZhizhin: https://github.com/hufrea/byedpi/blob/main/desync.c#L99

x/sockopt/is_sending_bytes_not_implemented.go
PeterZhizhin: Should be handy for Mac/iOS for disorder strategy. But it needs a separate implementation, based on `tcp_connection_info` struct: https://developer.apple.com/documentation/kernel/tcp_connection_info

x/sockopt/is_sending_bytes_linux.go
```diff
@@ -0,0 +1,42 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	// 1 == TCP_ESTABLISHED, but for some reason not available in the package
+	if tcpInfo.State != 1 {
```
fortuna, line 18: I believe you can use `unix.BPF_TCP_ESTABLISHED`: 
https://github.com/torvalds/linux/blob/2d5404caa8c7bb5c4e0435f94b28834ae5456623/net/ipv4/tcp.c#L2879

x/sockopt/sockopt.go
```diff
@@ -50,15 +60,51 @@ var _ HasHopLimit = (*hopLimitOption)(nil)
...
+
+func (o *tcpOptions) SetWaitingDelay(delay time.Duration) {
+	o.waitingDelay = delay
+}
+
+func (o *tcpOptions) OsSupportsWaitingUntilBytesAreSent() bool {
+	return isConnectionSendingBytesImplemented()
+}
+
+func (o *tcpOptions) WaitUntilBytesAreSent() error {
```
fortuna, line 92: This functionality is too high-level to live here.

I'd say we just need the bytes not sent and state options.

Then you can implement the logic to wait completely separate.

x/sockopt/is_sending_bytes_linux.go
fortuna, line 18: Why is this relevant? Won't `tcpInfo.Notsent_bytes` be zero if the connection is not established?

Also, we won't have the conenction object if it's not established.

x/sockopt/sockopt.go
```diff
@@ -50,15 +60,51 @@ var _ HasHopLimit = (*hopLimitOption)(nil)
 
 // TCPOptions represents options for TCP connections.
 type TCPOptions interface {
+	HasWaitUntilBytesAreSent
```
fortuna, line 63: We can't have such high-level functionality here. As mentioned below, we should perhaps expose lower level functionality instead.

x/sockopt/sockopt.go
fortuna, line 1: I worry about having TCPOptions have options that are not cross-platform. I'm still trying to figure out what a good design for this stuff is, and we'll only know better when we try different things.

In favor of simpler APIs, for now, let's go with a simple standalone function `BytesToSend(Rawconn)` that lives in the `fake` code, since that's the only strategy that really needs it. It shouldn't be defined in platforms that don't support it.

fortuna (changes requested):

x/sockopt/is_sending_bytes_linux.go
derlaft: > Why is this relevant? Won't tcpInfo.Notsent_bytes be zero if the connection is not established?

I don't get this comment. This check is not about connection establishment

> Also, we won't have the conenction object if it's not established.

This function should not be called when the connection is not yet established.

PeterZhizhin:
Fixed some comments. For more fixes, I'd like more input.

Waiting for the socket to send all bytes is useful for pretty much all strategies.
- `split`, `tlsfrag`: makes it more robust to kernel reassembling packets in it's memory buffer. I have seen `split:1*1000` sending stuff in chunks of 1, 2, 8, 16 bytes instead of always 1 byte. Combining it with a strategy that always waits in between writes will help.
- `oob` and `disorder`: guarantees that the data will be sent out of order. byeDPI currently does it on Linux, but not Windows. It's going to be "best effort" here. The strategies will work on all platforms, but will be more robust on Linux and Darwin.
- `fake`: same as `oob` and `disorder`, but seems to be more important there, since the kernel will do more stuff (like reading a temporary file)

PeterZhizhin:
I've pushed a new strategy into this PR: `waitstream`.

This new strategy will call `Write`, and call the new `WaitUntilBytesAreSent` after it. Skipping the error if it's not implemented.

Baseline:
```
go run *.go -timeout 1000 -transport "override:host=104.16.208.90|split:10000*1" -method HEAD -v https://meduza.io/
```

WireShark screenshot:
изображение

Data got packaged together into 5 packets of 1, 2, 118, 110, 35 bytes

Using the strategy:
```
go run *.go -timeout 1000 -transport "override:host=104.16.208.90|waitstream|split:10000*1" -method HEAD -v https://meduza.io/
```

WireShark screenshot:
изображение
Wireshark shows that everything gets send 1 byte at a time, as expected

x/sockopt/is_sending_bytes_not_implemented.go
fortuna: I haven't had split issues on macOS, so we may not need it.

x/sockopt/is_sending_bytes_linux.go
fortuna: I'm saying we can remove the `State` test. It's not doing anything. `tcpInfo.Notsent_bytes` will be zero if it's not established.

x/wait_stream/writer.go
```diff
@@ -0,0 +1,49 @@
...
+		conn:       conn,
+		tcpOptions: tcpOptions,
+	}
+}
+
+func (w *waitStreamWriter) Write(data []byte) (written int, err error) {
+	written, err = w.conn.Write(data)
+
+	// This may not be implemented, so it's best effort really.
+	waitUntilBytesAreSentErr := w.tcpOptions.WaitUntilBytesAreSent()
```
fortuna, line 43: Can you move the `WaitUntilBytesAreSent` logic out of `sockopt` and into the `wait_stream` package instead? Then revert `sockopt`.

This strategy is so tied to the Linux TCP implementation that there's no point in trying to make it portable.

`NewWriter` will need to take a `TCPConn` as input instead.

fortuna (changes requested):
I like the `waitstream` approach! It's a nice way to compose. But let's decouple from the TCPOptions. That's intended for low-level portable options.

x/wait_stream/writer.go
fortuna, line 43: I think this should be before the Write is done.

Also, this may significantly slow down communication. What's the performance impact? We will need to understand that for practical use.

We may want to restrict the wait for only when it's needed (on splits).

PeterZhizhin: Used a `fetch-speed` tool:
```
go run *.go -timeout 100500s -transport "waitstream"  http://speedtest.belwue.net/100M
Downloaded 100.00 MiB in 9.01s
Downloaded Speed: 11.10 MiB/s
```

Baseline:
```
go run *.go -timeout 100500s -transport ""  http://speedtest.belwue.net/100M                                                                                          
Downloaded 100.00 MiB in 9.75s
Downloaded Speed: 10.26 MiB/s
```

x/sockopt/is_sending_bytes_linux.go
derlaft: Got it, thanks

derlaft: Hmm, what about the cases when TCP connection is closing? This way we can abort waiting loop earlier. Maybe that's why it was there in the original code.

PR #372, opened 2025-02-11
Fixed double mutex acquisition deadlock when recreating UDP session

### LINKED ISSUE #342: Tunnel fails to close after Linux system resumes from sleep
On a Linux-based system, after resuming from sleep, it is impossible to close the tunnel without killing the process.  

Here’s the normal behavior when closing the tunnel:  
```
sudo go run github.com/Jigsaw-Code/outline-sdk/x/examples/outline-cli@latest -transport "ss://token"
OutlineVPN CLI (experimental)
[INFO] 2024/12/06 16:00:09 updated global IPv6 support: false
[INFO] 2024/12/06 16:00:09 remote server supports UDP, we will delegate all UDP packets to it
[INFO] 2024/12/06 16:00:09 routing traffic configured...
^C[INFO] 2024/12/06 16:47:43 received interrupt, terminating...
[INFO] 2024/12/06 16:47:43 routing table cleaned up
[INFO] 2024/12/06 16:47:43 DNS config restored
[INFO] 2024/12/06 16:47:43 OutlineDevice -> tun stopped: 274279765 <nil>
[INFO] 2024/12/06 16:47:43 tun -> OutlineDevice stopped: 27566484 read tun: file already closed
```  

However, after resuming from sleep, attempting to close the tunnel results in infinite waiting:  
```
OutlineVPN CLI (experimental)
[INFO] 2024/12/06 00:23:58 updated global IPv6 support: false
[INFO] 2024/12/06 00:23:58 remote server supports UDP, we will delegate all UDP packets to it
[INFO] 2024/12/06 00:23:58 routing traffic configured...
^C[INFO] 2024/12/06 10:21:42 received interrupt, terminating...
[INFO] 2024/12/06 10:21:42 routing table cleaned up
[INFO] 2024/12/06 10:21:42 DNS config restored
[INFO] 2024/12/06 10:21:42 OutlineDevice -> tun stopped: 264549559 <nil>
signal: killed
```  

The only way to close the tunnel in this case is to kill the `outline-cli` process manually.  

The issue occurs in `app_linux.go`, specifically here:  
```go
go func() {
    defer trafficCopyWg.Done()
    written, err := io.Copy(ss, tun)
    logging.Info.Printf("tun -> OutlineDevice stopped: %v %v\n", written, err)
}()
```  

This section does not receive an EOF, preventing the connection from closing.  

I believe the issue lies in how `io.Copy` blocks indefinitely while waiting for data from the TUN device. This behavior becomes problematic after the system wakes up from sleep, as the TUN device does not signal EOF or return from the `Read` call.

alamar:
minor: Opportunistically clean-up stale route rule in outline-cli
closes #342

### DIFF
--- a/network/lwip2transport/udp.go
+++ b/network/lwip2transport/udp.go
@@ -55,13 +55,15 @@ func (h *udpHandler) ReceiveTo(tunConn lwip.UDPConn, data []byte, destAddr *net.
 
 	h.mu.Lock()
 	reqSender, ok := h.senders[laddr]
+	h.mu.Unlock()
 	if !ok {
 		if reqSender, err = h.newSession(tunConn); err != nil {
 			return
 		}
+		h.mu.Lock()
 		h.senders[laddr] = reqSender
+		h.mu.Unlock()
 	}
-	h.mu.Unlock()
 
 	_, err = reqSender.WriteTo(data, destAddr.AddrPort())
 	return
--- a/network/lwip2transport/udp.go
+++ b/network/lwip2transport/udp.go
@@ -55,6 +55,7 @@ func (h *udpHandler) ReceiveTo(tunConn lwip.UDPConn, data []byte, destAddr *net.
 
 	h.mu.Lock()
 	reqSender, ok := h.senders[laddr]
+	// TODO: address synchronization issues with h.senders[laddr] and the following `if` in the future
 	h.mu.Unlock()
 	if !ok {
 		if reqSender, err = h.newSession(tunConn); err != nil {
--- a/x/examples/outline-cli/routing_linux.go
+++ b/x/examples/outline-cli/routing_linux.go
@@ -113,6 +113,8 @@ func setupIpRule(svrIp string, routingTable, routingPriority int) error {
 	ipRule.Invert = true
 
 	if err := netlink.RuleAdd(ipRule); err != nil {
+		// assuming duplicate from previous run, just to make sure it does not stays stale forever in the routing table
+		defer cleanUpRule()
 		return fmt.Errorf("failed to add IP rule (table %v, dst %v): %w", ipRule.Table, ipRule.Dst, err)
 	}
 	logging.Info.Printf("ip rule 'from all not to %v via table %v' created\n", ipRule.Dst, ipRule.Table)


x/examples/outline-cli/routing_linux.go
```diff
@@ -113,6 +113,8 @@ func setupIpRule(svrIp string, routingTable, routingPriority int) error {
 	ipRule.Invert = true
 
 	if err := netlink.RuleAdd(ipRule); err != nil {
+                // assuming duplicate from previous run, just to make sure it does not stays stale forever in the routing table
```
jyyi1, line 116: Please replace the spaces with tabs. (Maybe running `go fmt` would help?)

x/examples/outline-cli/routing_linux.go
```diff
@@ -113,6 +113,8 @@ func setupIpRule(svrIp string, routingTable, routingPriority int) error {
...
+		defer cleanUpRule()
```
jyyi1, line 117: Do we actually need the `defer` call? Why not just calling `cleanUpRule()` directly?

network/lwip2transport/udp.go
```diff
@@ -55,13 +55,15 @@ func (h *udpHandler) ReceiveTo(tunConn lwip.UDPConn, data []byte, destAddr *net.
 
 	h.mu.Lock()
 	reqSender, ok := h.senders[laddr]
+	h.mu.Unlock()
 	if !ok {
 		if reqSender, err = h.newSession(tunConn); err != nil {
```
jyyi1, line 60: I would suggest adding a `h.mu.Unlock()` before returning. The goal is to protect both `h.senders[laddr]` read and `h.senders[laddr] = reqSender` write.

jyyi1 (changes requested):
Thanks a lot for the fixing. We are experiencing similar issues on other products as well, this is very helpful! I have added some comments, please take a look.

network/lwip2transport/udp.go
alamar: Unfortunately, this is not possible since `newSession` will do `h.mu.Lock()` as well, so if it's not unlocked by then, it will always deadlock. As I understand, this execution path will always cause deadlock in the current code base.

I understand there may be an issue regarding access of `h.senders[laddr]`, but fixing it requires uprooting way more code that I understand at the moment. So I suggest to fix the hard deadlock issue now with minimal change - if there are any issues with UDP sessions access synchronization in the future, they may be taken care of separately. After all, this piece of code was essentially a dead end so there may be more bugs there.

alamar:
@jyyi1 please re-review - I've addressed formatting issue and replied on comments.

network/lwip2transport/udp.go
alamar: Perhaps I didn't understand you well and perhaps I didn't explain the issue well, so it's not possible to always do all `h.senders[laddr]` access inside the same mutex as they are interleaved with `newSession()` sometimes which requires mutex to be unlocked first.

jyyi1: Thanks for the explanation. It's acceptable to keep your implementation. Would you mind adding a TODO comment here to mention that we need a `newSessionUnlock` for better synchronization?

network/lwip2transport/udp.go
```diff
@@ -55,13 +55,15 @@ func (h *udpHandler) ReceiveTo(tunConn lwip.UDPConn, data []byte, destAddr *net.
 
 	h.mu.Lock()
 	reqSender, ok := h.senders[laddr]
+	h.mu.Unlock()
```
jyyi1, line 58:
```suggestion
	// TODO: address synchronization issues with h.senders[laddr] and the following `if` in the future
	h.mu.Unlock()
```

jyyi1:
> @jyyi1 please re-review - I've addressed formatting issue and replied on comments.

Thanks @alamar , the code LGTM. Please add a TODO comment and we will be good to merge it.

alamar:
@jyyi1 that is now done - please approve.

jyyi1 (approved):
Thanks for the contribution!

PR #427, opened 2025-04-17
bug(x) fix psiphon wrapper context

ohnorobo:
Use two contexts for psiphon so we don't cancel the tunnel if the input context is cancelled

Testing:
added the ability to run integration tests with a secret config in https://github.com/Jigsaw-Code/outline-sdk/pull/430


fortuna (approved):
Looks good. But we need to fix the tests

PR #438, opened 2025-04-24
fix(ww): update sdk mobileproxy version

daniellacosse:


x/examples/website-wrapper-app/README.md
```diff
@@ -29,6 +29,7 @@ To verify that your system has the necessary dependencies to generate your web w
 [Please refer to CapacitorJS's environment setup guide](https://capacitorjs.com/docs/getting-started/environment-setup#ios-requirements) and run `./doctor` to check to see if you have all the required dependencies.
 
 ```sh
+# make sure to run `npm run reset` between each build!
```
fortuna, line 32: Why do you need that?

x/examples/website-wrapper-app/.scripts/build_mobileproxy.sh
```diff
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 PLATFORM="$1"
-TAG="${2:-"x/v0.0.1"}"
+TAG="${2:-"x/v0.0.3"}"
```
fortuna, line 18: There's a better way. Create a tools.go file instead, like this:
https://github.com/Jigsaw-Code/outline-sdk/blob/main/x/mobileproxy/tools.go

Then go mod tidy will create the go.sum for you

x/examples/website-wrapper-app/tools.go
```diff
@@ -0,0 +1,5 @@
+package tools
```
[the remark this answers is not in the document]
fortuna: You need all the go dependencies you use. Need to add Go Mobile.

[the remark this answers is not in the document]
fortuna: Please run go mod tidy too and include the sum file.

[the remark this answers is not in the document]
fortuna: Wait, you don't have a separate module for the app wrapper, so you shouldn't be specifying any version at all. The x module already pulls everything you need.

x/examples/website-wrapper-app/.scripts/build_mobileproxy.sh
```diff
@@ -15,29 +15,24 @@
 # limitations under the License.
 
 PLATFORM="$1"
-TAG="${2:-"x/v0.0.1"}"
-OUTPUT="${3:-output}"
+OUTPUT="${2:-output}"
```
fortuna, line 18: I'd do it this way, so one can set the OUTPUT anywhere they want. It also cleans up all the references of OUTPUT in your code:
```suggestion
OUTPUT="${2:-$(pwd)/output}"
```

fortuna (approved):
Nice, this is cleaner. I have a suggestion for the OUTPUT variable.

x/examples/website-wrapper-app/.scripts/build_mobileproxy.sh
```diff
@@ -15,29 +15,24 @@
...
 
 if [[ "$OUTPUT" = "/" ]] || [[ "$OUTPUT" = "*" ]]; then
   echo "Error: OUTPUT cannot be '/' or '*'. These are dangerous values."
   exit 1
 fi
 
-git clone --depth 1 --branch "$TAG" https://github.com/Jigsaw-Code/outline-sdk.git "$OUTPUT/outline-sdk"
-cd "$OUTPUT/outline-sdk/x" || exit
-go build -o "$(pwd)/out/" golang.org/x/mobile/cmd/gomobile golang.org/x/mobile/cmd/gobind
+mkdir -p "$(pwd)/$OUTPUT/mobileproxy"
```
fortuna, line 25: You need to remove `$(pwd)` from all the references, that was the point.

fortuna (changes requested):

x/examples/website-wrapper-app/.scripts/build_mobileproxy.sh
```diff
@@ -15,29 +15,24 @@
...
 fi
 
-git clone --depth 1 --branch "$TAG" https://github.com/Jigsaw-Code/outline-sdk.git "$OUTPUT/outline-sdk"
-cd "$OUTPUT/outline-sdk/x" || exit
-go build -o "$(pwd)/out/" golang.org/x/mobile/cmd/gomobile golang.org/x/mobile/cmd/gobind
+mkdir -p "$OUTPUT/mobileproxy"
+
+go build -o "$OUTPUT/mobileproxy" golang.org/x/mobile/cmd/gomobile golang.org/x/mobile/cmd/gobind
 
 if [ "$PLATFORM" = "ios" ]; then
```
fortuna, line 29: Not needed for this PR, but you should use the `[[...]]` tests instead as per the style guide. It's safer

x/examples/website-wrapper-app/tools.go
fortuna, line 1: Revert. This is no longer needed, since x already brings the tools for mobileproxy. We'll need that if we move to a separate module though.

x/examples/website-wrapper-app/tools.go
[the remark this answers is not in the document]
fortuna: If you need them to be versioned separately, you will need to define a new module. But I don't think we need that at this point.

That breakage would probably leak into the wrapper anyway on upgrade, even on separate modules, since we don't have tests.

fortuna (changes requested):

fortuna (approved):

PR #475, opened 2025-07-07
website-wrapper-app: Restore open:* scripts (removed in #458)

GrantASL19:
These were removed in #458:

https://github.com/Jigsaw-Code/outline-sdk/pull/458/files#diff-c2486eb7d0eaea9be3699c240db3784f1f1548d8c40b560f1ee5fa1f2fbe26c0L21-L22

I assume this was because the directory name of the project became dynamic.

This change replaces the previous `cd output/wrapper_app_project` with `cd $(find output -mindepth 1 -maxdepth 1 -type d ! -name 'mobileproxy' | head -n 1)`, which is essentially "open the first subdirectory of output not named "mobileproxy".

I don’t think there’s a good way to avoid the repetition but please let me know if you have any ideas.

### DIFF
--- a/x/examples/website-wrapper-app/package.json
+++ b/x/examples/website-wrapper-app/package.json
@@ -19,6 +19,8 @@
     "build": "node ./wrapper_app_project/.scripts/build.mjs",
     "clean": "npx rimraf node_modules/ output/",
     "doctor": "./doctor",
+    "open:android": "cd $(find output -mindepth 1 -maxdepth 1 -type d ! -name 'mobileproxy' | head -n 1) && npm run open:android",
+    "open:ios": "cd $(find output -mindepth 1 -maxdepth 1 -type d ! -name 'mobileproxy' | head -n 1) && npm run open:ios",
     "reset": "npm run clean && npm ci",
     "start:navigator": "node ./basic_navigator_example/.scripts/start.mjs"
   },


x/examples/website-wrapper-app/package.json
```diff
@@ -19,6 +19,8 @@
     "build": "node ./wrapper_app_project/.scripts/build.mjs",
     "clean": "npx rimraf node_modules/ output/",
     "doctor": "./doctor",
+    "open:android": "cd $(find output -mindepth 1 -maxdepth 1 -type d ! -name 'mobileproxy' | head -n 1) && npm run open:android",
```
daniellacosse, line 22: Does `npx cap open {ios,android}` not work? I think we can simplify it to that.

daniellacosse: https://capacitorjs.com/docs/cli/commands/open

GrantASL19: > Does `npx cap open {ios,android}` not work? I think we can simplify it to that.

The [template file](https://github.com/Jigsaw-Code/outline-sdk/blob/main/x/examples/website-wrapper-app/wrapper_app_project/template/package.json) has the "open:ios" and "open:android" scripts which the [previous implementation](https://github.com/Jigsaw-Code/outline-sdk/pull/458/files#diff-c2486eb7d0eaea9be3699c240db3784f1f1548d8c40b560f1ee5fa1f2fbe26c0L21-L22) used, so I figured I should replicate that (with the fix for the dynamic app directory name).

Alternatively I can get rid of those scripts and just call `cap open {ios,android}` without `npx` since `@capacitor/cli` is installed in the app directory.

PR #497, opened 2025-08-06
fix(x/websocket): concurrent read/write panics

jyyi1:
From the [`websocket` package's document](https://pkg.go.dev/github.com/gorilla/websocket#hdr-Concurrency):

> Applications are responsible for ensuring that no more than one goroutine calls the write methods (NextWriter, SetWriteDeadline, WriteMessage, WriteJSON, EnableWriteCompression, SetCompressionLevel) concurrently and that no more than one goroutine calls the read methods (NextReader, SetReadDeadline, ReadMessage, ReadJSON, SetPongHandler, SetPingHandler) concurrently.

This will potentially fix this Outline Client issue: https://github.com/Jigsaw-Code/outline-apps/issues/2573

### DIFF
--- a/x/websocket/endpoint.go
+++ b/x/websocket/endpoint.go
@@ -26,6 +26,7 @@ import (
 	"net/url"
 	"runtime"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/Jigsaw-Code/outline-sdk/transport"
@@ -111,7 +112,11 @@ func newGorillaConn(wsConn *websocket.Conn) *gorillaConn {
 }
 
 type gorillaConn struct {
-	wsConn        *websocket.Conn
+	wsConn *websocket.Conn
+
+	// websocket.Conn is not safe for concurrent use
+	readMu, writeMu sync.Mutex
+
 	writeErr      error
 	readErr       error
 	pendingReader io.Reader
@@ -140,6 +145,9 @@ func (c *gorillaConn) SetWriteDeadline(deadline time.Time) error {
 }
 
 func (c *gorillaConn) Read(buf []byte) (int, error) {
+	c.readMu.Lock()
+	defer c.readMu.Unlock()
+
 	if c.readErr != nil {
 		return 0, c.readErr
 	}
@@ -177,23 +185,32 @@ func (c *gorillaConn) Read(buf []byte) (int, error) {
 }
 
 func (c *gorillaConn) Write(buf []byte) (int, error) {
+	c.writeMu.Lock()
+	defer c.writeMu.Unlock()
+
+	if c.writeErr != nil {
+		return 0, c.writeErr
+	}
 	err := c.wsConn.WriteMessage(websocket.BinaryMessage, buf)
 	if err != nil {
-		if c.writeErr != nil {
-			return 0, c.writeErr
-		}
 		return 0, err
 	}
 	return len(buf), nil
 }
 
 func (c *gorillaConn) CloseRead() error {
+	c.readMu.Lock()
+	defer c.readMu.Unlock()
+
 	c.readErr = net.ErrClosed
 	c.wsConn.SetReadDeadline(time.Now())
 	return nil
 }
 
 func (c *gorillaConn) CloseWrite() error {
+	c.writeMu.Lock()
+	defer c.writeMu.Unlock()
+
 	// Send close message.
 	message := websocket.FormatCloseMessage(websocket.CloseNormalClosure, "")
 	c.wsConn.WriteControl(websocket.CloseMessage, message, time.Now().Add(time.Second))
--- a/x/websocket/endpoint_test.go
+++ b/x/websocket/endpoint_test.go
@@ -15,11 +15,17 @@
 package websocket
 
 import (
+	"bufio"
 	"context"
 	"errors"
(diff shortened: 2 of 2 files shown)


x/websocket/endpoint.go
```diff
@@ -177,23 +185,32 @@ func (c *gorillaConn) Read(buf []byte) (int, error) {
...
 func (c *gorillaConn) Write(buf []byte) (int, error) {
+	c.writeMu.Lock()
+	defer c.writeMu.Unlock()
+
+	if c.writeErr != nil {
+		return 0, c.writeErr
+	}
 	err := c.wsConn.WriteMessage(websocket.BinaryMessage, buf)
 	if err != nil {
-		if c.writeErr != nil {
```
fortuna, line 182: We still need this if the connection is closed during the write.

x/websocket/endpoint.go
```diff
@@ -177,23 +185,32 @@ func (c *gorillaConn) Read(buf []byte) (int, error) {
...
+	c.readMu.Lock()
+	defer c.readMu.Unlock()
+
 	c.readErr = net.ErrClosed
 	c.wsConn.SetReadDeadline(time.Now())
 	return nil
 }
 
 func (c *gorillaConn) CloseWrite() error {
+	c.writeMu.Lock()
```
fortuna: We need to be able to close during a write.

x/websocket/endpoint.go
```diff
@@ -177,23 +185,32 @@ func (c *gorillaConn) Read(buf []byte) (int, error) {
...
-		if c.writeErr != nil {
-			return 0, c.writeErr
-		}
 		return 0, err
 	}
 	return len(buf), nil
 }
 
 func (c *gorillaConn) CloseRead() error {
+	c.readMu.Lock()
```
fortuna: We need to be able to close during a read.

x/websocket/endpoint.go
jyyi1: Reverted

x/websocket/endpoint.go
jyyi1: Reverted

x/websocket/endpoint.go
jyyi1: Reverted

x/websocket/endpoint.go
```diff
@@ -111,7 +112,11 @@ func newGorillaConn(wsConn *websocket.Conn) *gorillaConn {
 }
 
 type gorillaConn struct {
-	wsConn        *websocket.Conn
+	wsConn *websocket.Conn
+
+	// websocket.Conn is not safe for concurrent use
```
fortuna, line 117: Perhaps add a comment mentioning the bug for future context.

fortuna (approved):
Thanks for the fix!

jyyi1:
> Thanks for the fix!

The [CI catches the potential race condition between `Read` and `CloseRead`](https://github.com/Jigsaw-Code/outline-sdk/actions/runs/16789879025/job/47549286365?pr=497#step:12:41). Though I can fix the test by synchronizing these two calls, the risk of this issue still exists.

jyyi1:
> Perhaps a fix is to change CloseWrite to first set the deadline to now (to cancel existing writes), acquire the lock, reset the deadline, write the CloseMessage, reset the deadline and release the lock.

Unfortunately this approach won't work. The [`gorilla/websocket` document](https://pkg.go.dev/github.com/gorilla/websocket#hdr-Concurrency) clearly states that:  "no more than one goroutine calls the write methods (`SetWriteDeadline`, `Write`, ...) and the read method (`SetReadDeadline`, `Read`, ...)"; a race condition would still exist and caught by the Go test framework.

I've fixed the "Concurrent Read" test that was failing due to concurrent `Read` and `CloseRead`. I also added two new tests for `CloseWrite` and `CloseRead` concurrency but had to skip them with `t.Skip()` since the issue hasn't been fixed yet.

x/websocket/endpoint.go
jyyi1: Added

fortuna:
I think my suggestion was misunderstood. This is the sequence:

1.  CloseWrite sets the deadline to now (to cancel existing writes) and requests the write lock
2. Pending write returns (with deadline exceeded) and releases the lock
3. CloseWrite acquires the lock, removes the deadline, writes the CloseMessage and releases the lock.

There's no concurrent writes there.

jyyi1:
> I think my suggestion was misunderstood. This is the sequence:
> 
> 1. CloseWrite sets the deadline to now (to cancel existing writes) and requests the write lock
> 2. Pending write returns (with deadline exceeded) and releases the lock
> 3. CloseWrite acquires the lock, removes the deadline, writes the CloseMessage and releases the lock.
> 
> There's no concurrent writes there.
> 
> It's possible that a second write will start and acquire the lock before CloseWrite, but all writes will fail because of the deadline. Eventually the CloseWrite will acquire the lock.

You mean?

```
goroutine1: a pending Write()
goroutine2: SetWriteDeadline(now)
```

These two calls are happening concurrently: `Write` and `SetWriteDeadline`. And it is forbidden.

fortuna (approved):

PR #503, opened 2025-08-21
Update golang crypto library

ohnorobo:
run `go get -u golang.org/x/crypto` in sdk and in /x

This is to fix the error

```
E  panic: tls: ConnectionState is not equal to tls.ConnectionState: struct field count mismatch: 17 vs 16
```

Someone was seeing this while trying to use mobileproxy in an android app. Due to a [new field being added](https://github.com/golang/go/blame/9de69f6913dc253ee8ce67043f543bedde28fdb4/src/crypto/tls/common.go#L315) in crypto/tls Conn.ConnectionState

### DIFF
--- a/x/go.mod
+++ b/x/go.mod
@@ -1,5 +1,7 @@
 module github.com/Jigsaw-Code/outline-sdk/x
 
+go 1.23.0
+
 require (
 	github.com/Jigsaw-Code/outline-sdk v0.0.20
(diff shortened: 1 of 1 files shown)


fortuna (approved):
I think it's good to upgrade the dependencies, but the explanation does not make sense to me. We build a static binary, why would there be any conflict?

Were you able to reproduce the bug? Does this fix it?

(Also, kind of crazy how many new libraries it's pulling)

PR #505, opened 2025-08-21
fix(x/mobileproxy): prevent panic when StrategyCache is not set

### LINKED ISSUE #501: mobileproxy segfaults if strategy cache is not set
Issue:

Building mobileproxy like so
```
go build -tags psiphon -o "$(pwd)/out/" golang.org/x/mobile/cmd/gomobile http://golang.org/x/mobile/cmd/gobind
PATH="$(pwd)/out:$PATH" gomobile bind -ldflags='-s -w' -target=android -androidapi=21 -o "$(pwd)/out/mobileproxy.aar" github.com/Jigsaw-Code/outline-sdk/x/mobileproxy
```

and then using the resulting .aar file as a proxy in another app leads to a segfault
```
2025-08-20 15:53:33.997 21973-0     Go                      [redacted]                         E  panic: runtime error: invalid memory address or nil pointer dereference
2025-08-20 15:53:33.997 21973-0     Go                      [redacted]                         E  [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x786f28a07c]
```

seemingly around
```
Seq.java
    public static int incRef(Object o) {
        return tracker.inc(o);
    }
```

Eventually it was found that setting a StrategyCache in the options as is done [here](https://github.com/Jigsaw-Code/outline-app-maker/blob/8f760716cc72e0a7c5bb0fe2892156dc21db4cfe/wrapper_app_project/template/android/app/src/main/java/org/getoutline/pwa/MainActivity.kt.handlebars#L42) stopped the crash from happening. Removing that line from outline-app-maker also causes it to crash.

In our [documentation](https://pkg.go.dev/github.com/Jigsaw-Code/outline-sdk/x@v0.0.6/mobileproxy#SmartDialerOptions.SetStrategyCache) we describe the StrategyCache as optional. Either we should not crash or we should make it required with a clearer error.

jyyi1:
When `SmartDialerOptions.SetStrategyCache` is not called, `opt.cache` is `nil`. This `nil` value of type `*strategyCacheAdapter` is then assigned to the `Cache` field of `smart.StrategyFinder`, which is an interface.

This results in a non-nil interface with a nil value, which causes a panic when its methods are called inside `finder.NewDialer`.

This change fixes the issue by ensuring that a `nil` interface is passed to `smart.StrategyFinder` when `opt.cache` is `nil`, preventing the panic.

> Note that `LogWriter` doesn't need the same `if` check because `logWriter` is already an interface, so a nil `logWriter` is a true `nil` interface, which is safe to assign directly.

Fixes #501

### DIFF
--- a/x/mobileproxy/stream_dialer.go
+++ b/x/mobileproxy/stream_dialer.go
@@ -101,7 +101,11 @@ func (opt *SmartDialerOptions) NewStreamDialer() (*StreamDialer, error) {
 		TestTimeout:  opt.testTimeout,
 		StreamDialer: opt.baseSD,
 		PacketDialer: opt.basePD,
-		Cache:        opt.cache,
+	}
+	// When assigning a nil concrete value to an interface, the interface is not nil
+	// but contains a nil value. The `if` check is necessary to avoid a panic.
+	if opt.cache != nil {
+		finder.Cache = opt.cache
 	}
 
 	dialer, err := finder.NewDialer(context.Background(), opt.testDomains, opt.config)


fortuna (approved):
Thanks for the quick fix. A test would be helpful too.

PR #539, opened 2025-10-15
feat(x): parallelize smartdialer testing when given multiple test domains

ohnorobo:
We don't parallelize the testing on multiple  test domains for the smartdialer, which can be slow if some of the domains take a longer time.

This adds another layer of parallelization for the test domains. (on top of using racetest for the strategies)

Some example testing output with a real config: https://docs.google.com/document/d/1kS7aeEe4UrXabOskd2B0QzR3QzmjDj26QabXnpeJW7I/edit?resourcekey=0-N2f-VSYcs1rBx3hG6s38OA&tab=t.0


fortuna (approved):
We had issues with the backend throttling, showing captcha because of too many quick requests, but since we are talking to different backends in parallel, this will probably be ok.

Thanks for the change and the extra context.

PR #559, opened 2025-12-15
ci: free up disk space to prevent android emulator creation failure

jyyi1:
The `android-test` job was failing with ["Not enough space to create userdata partition"](https://github.com/Jigsaw-Code/outline-sdk/actions/runs/20183073943/job/58094204027?pr=556#step:7:98) in #556 . This PR adds a step to remove unused software from the runner to free up disk space and prevent this failure.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -84,6 +84,14 @@ jobs:
           distribution: "zulu"
           java-version: "17"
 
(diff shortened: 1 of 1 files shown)


jyyi1:
> This is crazy. How are we running out of disk space????

GitHub pre-installed many libraries. For example, the ones we removed in this PR take 10GB+ space.

PR #594, opened 2026-04-01
feat: Introduce dns.RawResolver interface

fortuna:
# Motivation

The main goal here is to decouple this foundational piece from the DNS parsing library. I've had a few issues with `dns.Resolver` while working on ECH research:

* dnsmessage fails to parse DNS labels with dots in it, which is used in the email field of SOA records.
* dnsmessage may not support the latest record types. In my case, it didn't support the HTTPS and SVCB records.
 
I ended up forced to use miekg's library instead of the SDK for that work.

This decoupling will allow the user to use whatever parsing library they prefer.

It also streamlines the implementation of the system resolver, removing unnecessary transformations when we fetch a `[]byte` response from the system APIs. See https://github.com/OutlineFoundation/outline-sdk/pull/592 for where I'm going.

The new interface takes the output buffer as input, which lets the caller control the lifecycle of the buffer and avoid unnecessary allocations.

I also want this to be a building block for the [dnstt implementation](https://github.com/OutlineFoundation/outline-sdk/discussions/574).

# Changes

The new API includes:
* `RawResolver` (takes QNAME + QTYPE + output buffer)
* `FuncRawResolver` - convenience `func` -> `RawResolver` adaptor
* `RawToResolver` - convenience adaptor to map to the current interface.

### DIFF
--- a/dns/doc.go
+++ b/dns/doc.go
@@ -45,5 +45,13 @@ in parallel, as per the [Happy Eyeballs v2] algorithm.
 [DNS-over-TLS]: https://datatracker.ietf.org/doc/html/rfc7858
 [DNS-over-HTTPS]: https://datatracker.ietf.org/doc/html/rfc8484
 [Happy Eyeballs v2]: https://datatracker.ietf.org/doc/html/rfc8305
+
+# RawResolver
+
+The [RawResolver] interface provides a lower-level primitive operating directly on wire-format byte arrays and primitive types (domain name, query type, and an output buffer). This design serves three main purposes:
+
+ 1. Zero-dependency footprint: Bypassing rigid DNS parsing libraries allows Outline to pack esoteric labels natively natively (e.g. obscure DNS tunneling headers).
+ 2. Future-proof compatibility: Callers can inject or parse the raw responses using any external library, including those supporting newer record types.
+ 3. Zero-allocation cycles: Providing a reusable `buf []byte` hands memory management to the caller, removing slice allocations in the hot path.
 */
 package dns
--- a/dns/resolver.go
+++ b/dns/resolver.go
@@ -68,6 +68,40 @@ func (f FuncResolver) Query(ctx context.Context, q dnsmessage.Question) (*dnsmes
 	return f(ctx, q)
 }
 
+// RawResolver can query DNS and return the raw wire-format response bytes as defined in RFC 1035.
+// Using plain name and qtype avoids a dependency on any specific DNS parsing library,
+// allowing callers to parse the response with any library — including those that support
+// record types not yet recognized by golang.org/x/net/dns/dnsmessage.
+type RawResolver interface {
+	QueryRaw(ctx context.Context, name string, qtype uint16) ([]byte, error)
+}
+
+// FuncRawResolver is a [RawResolver] that uses the given function to query DNS.
+type FuncRawResolver func(ctx context.Context, name string, qtype uint16) ([]byte, error)
+
+// QueryRaw implements the [RawResolver] interface.
+func (f FuncRawResolver) QueryRaw(ctx context.Context, name string, qtype uint16) ([]byte, error) {
+	return f(ctx, name, qtype)
+}
+
+// RawToResolver wraps a [RawResolver] in a [Resolver] that parses the wire-format
+// response bytes using golang.org/x/net/dns/dnsmessage.
+// The underlying [RawResolver] is responsible for ID matching and returning valid bytes;
+// this adapter only unpacks the result.
+func RawToResolver(r RawResolver) Resolver {
+	return FuncResolver(func(ctx context.Context, q dnsmessage.Question) (*dnsmessage.Message, error) {
+		raw, err := r.QueryRaw(ctx, q.Name.String(), uint16(q.Type))
+		if err != nil {
+			return nil, err
+		}
+		var msg dnsmessage.Message
+		if err := msg.Unpack(raw); err != nil {
+			return nil, &nestedError{ErrBadResponse, fmt.Errorf("failed to unpack DNS response: %w", err)}
+		}
+		return &msg, nil
+	})
+}
+
 // NewQuestion is a convenience function to create a [dnsmessage.Question].
 // The input domain is interpreted as fully-qualified. If the end "." is missing, it's added.
 func NewQuestion(domain string, qtype dnsmessage.Type) (*dnsmessage.Question, error) {
@@ -92,6 +126,15 @@ func NewQuestion(domain string, qtype dnsmessage.Type) (*dnsmessage.Question, er
 // This is based on an MTU of 1280, which is required by the IPv6 specification, minus 48 bytes
 // for the IPv6 and UDP headers".
 const maxUDPMessageSize = 1232
+
+// makeQuestion constructs a dnsmessage.Question from a plain name and record type.
+func makeQuestion(name string, qtype uint16) (dnsmessage.Question, error) {
+	q, err := NewQuestion(name, dnsmessage.Type(qtype))
+	if err != nil {
+		return dnsmessage.Question{}, err
+	}
+	return *q, nil
+}
 
 // appendRequest appends the bytes a DNS request using the id and question to buf.
 func appendRequest(id uint16, q dnsmessage.Question, buf []byte) ([]byte, error) {
@@ -167,8 +210,13 @@ func checkResponse(reqID uint16, reqQues dnsmessage.Question, respHdr dnsmessage
 }
 
 // queryDatagram implements a DNS query over a datagram protocol.
-func queryDatagram(conn io.ReadWriter, q dnsmessage.Question) (*dnsmessage.Message, error) {
+// It validates the response ID and question echo before returning raw wire-format bytes.
+func queryDatagram(conn io.ReadWriter, name string, qtype uint16) ([]byte, error) {
 	// Reference: https://cs.opensource.google/go/go/+/master:src/net/dnsclient_unix.go?q=func:dnsPacketRoundTrip&ss=go%2Fgo
+	q, err := makeQuestion(name, qtype)
+	if err != nil {
+		return nil, &nestedError{ErrBadRequest, fmt.Errorf("invalid question: %w", err)}
+	}
 	id := uint16(rand.Uint32())
 	buf, err := appendRequest(id, q, make([]byte, 0, maxUDPMessageSize))
 	if err != nil {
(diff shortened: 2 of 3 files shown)


dns/resolver.go
```diff
@@ -68,6 +68,40 @@ func (f FuncResolver) Query(ctx context.Context, q dnsmessage.Question) (*dnsmes
 	return f(ctx, q)
 }
 
+// RawResolver can query DNS and return the raw wire-format response bytes as defined in RFC 1035.
+// Using plain name and qtype avoids a dependency on any specific DNS parsing library,
+// allowing callers to parse the response with any library — including those that support
+// record types not yet recognized by golang.org/x/net/dns/dnsmessage.
+type RawResolver interface {
+	QueryRaw(ctx context.Context, name string, qtype uint16) ([]byte, error)
```
fortuna: We need to decide on the qname format. Could be the native `[]byte` format with length-prefixed labels, or `[][]byte` (slice of labels), or a escaped string ("foo\.bar.com").

While this doesn't affect regular usage, it can help use cases like DNS tunneling.

dns/resolver.go
```diff
@@ -93,6 +127,15 @@ func NewQuestion(domain string, qtype dnsmessage.Type) (*dnsmessage.Question, er
 // for the IPv6 and UDP headers".
 const maxUDPMessageSize = 1232
 
+// makeQuestion constructs a dnsmessage.Question from a plain name and record type.
+func makeQuestion(name string, qtype uint16) (dnsmessage.Question, error) {
+	q, err := NewQuestion(name, dnsmessage.Type(qtype))
```
fortuna: Should we just craft the binary format here instead of using the dependency? Not sure how it handles obscure labels.

dns/resolver.go
fortuna: Should we pass the buffer into the query? That allows that caller to control its lifecycle (like using a buffer pool) and potentially avoid or reduce allocations. The response buffer can be used for the request too. 

This solve the problem of setting the max size in the EDNS settings in the OPT record for UDP.
One challenge is that, for streams, we may want to dynamically allocate the buffer based on the packet length.

Do we need an interface like in Websockets, where you can have messages that are read over multiple reads? We could return a Reader.

fortuna (commented):
Some API questions to consider

dns/resolver.go
```diff
@@ -68,6 +68,41 @@ func (f FuncResolver) Query(ctx context.Context, q dnsmessage.Question) (*dnsmes
...
+	return f(ctx, name, qtype, buf)
+}
+
+// RawToResolver wraps a [RawResolver] in a [Resolver] that parses the wire-format
+// response bytes using golang.org/x/net/dns/dnsmessage.
+// The underlying [RawResolver] is responsible for ID matching and returning valid bytes;
+// this adapter only unpacks the result.
+func RawToResolver(r RawResolver) Resolver {
+	return FuncResolver(func(ctx context.Context, q dnsmessage.Question) (*dnsmessage.Message, error) {
+		buf := make([]byte, 0, maxUDPMessageSize)
```
fortuna, line 93: What should the size be? I'm not convinced this is the best approach.

dns/resolver.go
fortuna: Should we read it into buf instead to avoid allocating another buffer?

dns/doc.go
```diff
@@ -45,5 +45,13 @@ in parallel, as per the [Happy Eyeballs v2] algorithm.
...
 [DNS-over-HTTPS]: https://datatracker.ietf.org/doc/html/rfc8484
 [Happy Eyeballs v2]: https://datatracker.ietf.org/doc/html/rfc8305
+
+# RawResolver
+
+The [RawResolver] interface provides a lower-level primitive operating directly on wire-format byte arrays and primitive types (domain name, query type, and an output buffer). This design serves three main purposes:
+
+ 1. Zero-dependency footprint: Bypassing rigid DNS parsing libraries allows Outline to pack esoteric labels natively natively (e.g. obscure DNS tunneling headers).
+ 2. Future-proof compatibility: Callers can inject or parse the raw responses using any external library, including those supporting newer record types.
+ 3. Zero-allocation cycles: Providing a reusable `buf []byte` hands memory management to the caller, removing slice allocations in the hot path.
```
fortuna, line 55: Let's reorder:
1. Future proof
2. Zero Allocation
3. BYO parsing

dns/resolver.go
```diff
@@ -68,6 +68,41 @@ func (f FuncResolver) Query(ctx context.Context, q dnsmessage.Question) (*dnsmes
 	return f(ctx, q)
 }
 
+// RawResolver can query DNS and return the raw wire-format response bytes as defined in RFC 1035.
+// Using plain name and qtype avoids a dependency on any specific DNS parsing library,
+// allowing callers to parse the response with any library — including those that support
+// record types not yet recognized by golang.org/x/net/dns/dnsmessage.
+type RawResolver interface {
+	QueryRaw(ctx context.Context, name string, qtype uint16, buf []byte) ([]byte, error)
```
fortuna, line 76: We should explain that the returned buffer may or may not be a slice of the input. It depends on the implementation.

Or should we change that to guarantee it?

The main challenge is when we have dynamic lengths, as in TCP and HTTP

dns/resolver.go
```diff
@@ -68,6 +68,41 @@ func (f FuncResolver) Query(ctx context.Context, q dnsmessage.Question) (*dnsmes
...
+type FuncRawResolver func(ctx context.Context, name string, qtype uint16, buf []byte) ([]byte, error)
+
+// QueryRaw implements the [RawResolver] interface.
+func (f FuncRawResolver) QueryRaw(ctx context.Context, name string, qtype uint16, buf []byte) ([]byte, error) {
+	return f(ctx, name, qtype, buf)
+}
+
+// RawToResolver wraps a [RawResolver] in a [Resolver] that parses the wire-format
+// response bytes using golang.org/x/net/dns/dnsmessage.
+// The underlying [RawResolver] is responsible for ID matching and returning valid bytes;
```
fortuna, line 89: We can remove this line.

dns/resolver.go
fortuna, line 76: Is there a formal definition for the qname format used here with escaping?

dns/resolver.go
```diff
@@ -93,32 +128,96 @@ func NewQuestion(domain string, qtype dnsmessage.Type) (*dnsmessage.Question, er
...
-		return nil, fmt.Errorf("add question failed: %w", err)
-	}
-	if err := b.StartAdditionals(); err != nil {
-		return nil, fmt.Errorf("start additionals failed: %w", err)
+// appendName splits a dot-separated domain name into DNS wire-format labels and appends them to buf.
+// It supports escaping characters using a backslash.
+func appendName(buf []byte, name string) ([]byte, error) {
+	if len(name) > 0 && name[len(name)-1] == '.' {
+		// Only strip the last dot if it's not escaped
+		if len(name) < 2 || name[len(name)-2] != '\\' {
```
fortuna, line 136: This is broken if the string ends in `\\.`.
We need to unescape first, before removing the root dot.

dns/resolver.go
```diff
@@ -93,32 +128,96 @@ func NewQuestion(domain string, qtype dnsmessage.Type) (*dnsmessage.Question, er
...
+		return nil, errors.New("name too long")
 	}
+	return buf, nil
+}
 
-	buf, err := b.Finish()
+// appendRequestRaw constructs a complete wire-format DNS request (Header, Question, OPT) and appends it to buf.
+func appendRequestRaw(id uint16, name string, qtype uint16, buf []byte) ([]byte, error) {
+	udpSize := uint16(cap(buf))
+	if udpSize < 512 {
```
fortuna, line 187: Why is this? Is it defined in a RFC? We need a link.

dns/resolver.go
```diff
@@ -93,32 +128,96 @@ func NewQuestion(domain string, qtype dnsmessage.Type) (*dnsmessage.Question, er
...
+	if len(buf)-startLength > 255 {
+		return nil, errors.New("name too long")
 	}
+	return buf, nil
+}
 
-	buf, err := b.Finish()
+// appendRequestRaw constructs a complete wire-format DNS request (Header, Question, OPT) and appends it to buf.
+func appendRequestRaw(id uint16, name string, qtype uint16, buf []byte) ([]byte, error) {
+	udpSize := uint16(cap(buf))
```
fortuna, line 186: Rename to maxResponseSize.

dns/resolver.go
```diff
@@ -93,32 +128,96 @@ func NewQuestion(domain string, qtype dnsmessage.Type) (*dnsmessage.Question, er
...
-		return nil, fmt.Errorf("message serialization failed: %w", err)
-	}
+		return nil, fmt.Errorf("invalid question name: %w", err)
+	}
+	buf = append(buf,
+		byte(qtype>>8), byte(qtype), // QTYPE
+		0x00, 0x01, // QCLASS (INET=1)
+	)
+
+	// Additional Section (OPT)
```
fortuna, line 212: Does this make sense for Stream and HTTP?
What should be the length in those cases?

dns/resolver.go
fortuna: Found it: https://datatracker.ietf.org/doc/html/rfc1035#section-2.3.4

dns/resolver.go
fortuna: The OPT record should be set by the implementation.
UDP needs to increase the payload size.
TCP can set a keep alive
TLS may want to set padding
All of them may set the client subnet or the extended error (https://datatracker.ietf.org/doc/html/rfc8914)

Can we layer the subnet and extended error behavior?

dns/resolver.go
fortuna: Perhaps we should set the max message size in the resolver object. Each implementation will have different parameters.
We may want to specify OPT options.

Also see https://datatracker.ietf.org/doc/html/rfc6891#section-6.2.5

dns/resolver.go
```diff
@@ -144,36 +243,74 @@ func equalASCIIName(x, y dnsmessage.Name) bool {
...
+
+	reqQLen := len(expectedName) + 4
+	if len(rawResp) < 12+reqQLen {
+		return errors.New("response too short for echoed question")
 	}
+	
+	respQName := rawResp[12 : 12+len(expectedName)]
 
+	// Case-insensitive comparison, as the server may echo back randomized caps (e.g. 0x20 encoding)
+	for i := 0; i < len(expectedName); i++ {
```
fortuna, line 284: Move to a function. Also, need to compare the lengths!

dns/resolver.go
```diff
@@ -144,36 +243,74 @@ func equalASCIIName(x, y dnsmessage.Name) bool {
...
+		if foldCase(expectedName[i]) != foldCase(respQName[i]) {
+			return fmt.Errorf("response question name doesn't match request. Expected %x, got %x", expectedName, respQName)
+		}
+	}
+	
+	respQType := binary.BigEndian.Uint16(rawResp[12+len(expectedName):])
```
fortuna, line 290: We should check type and class first.

dns/resolver.go
```diff
@@ -144,36 +243,74 @@ func equalASCIIName(x, y dnsmessage.Name) bool {
 	return true
 }
 
-func checkResponse(reqID uint16, reqQues dnsmessage.Question, respHdr dnsmessage.Header, respQs []dnsmessage.Question) error {
-	if !respHdr.Response {
+// checkResponseRaw verifies the DNS response matches the request parameters directly on the wire bytes.
+func checkResponseRaw(reqID uint16, name string, qtype uint16, rawResp []byte) error {
```
fortuna, line 247: It seems like we may need another abstraction layer. This kind of check is standard, and independent on how we exchange messages (though the message exchanges depend on the message id)

There's also things like padding, extended errors, client subnet, DNSSEC, and so on.... Where should that live?

Oh, and caching!

dns/resolver.go
```diff
@@ -256,13 +396,13 @@ func ensurePort(address string, defaultPort string) string {
 	return address
 }
 
-// NewUDPResolver creates a [Resolver] that implements the DNS-over-UDP protocol, using a [transport.PacketDialer] for transport.
+// NewUDPRawResolver creates a [RawResolver] that implements the DNS-over-UDP protocol, using a [transport.PacketDialer] for transport.
 // It uses a different port for every request.
 //
 // [DNS-over-UDP]: https://datatracker.ietf.org/doc/html/rfc1035#section-4.2.1
-func NewUDPResolver(pd transport.PacketDialer, resolverAddr string) Resolver {
+func NewUDPRawResolver(pd transport.PacketDialer, resolverAddr string) RawResolver {
```
fortuna, line 403: I don't think these factory functions work well. We should consider a way to specify configs, like the max packet length for UDP, and layer other functionality that are common across implementations.

dns/resolver.go
```diff
@@ -375,19 +543,49 @@ func NewHTTPSResolver(sd transport.StreamDialer, resolverAddr string, url string
...
 		}
 
 		// Process response.
-		var msg dnsmessage.Message
-		if err = msg.Unpack(response); err != nil {
-			return nil, &nestedError{ErrBadResponse, fmt.Errorf("failed to unpack DNS response: %w", err)}
-		}
-		if err := checkResponse(0, q, msg.Header, msg.Questions); err != nil {
+		// Ignore invalid packets that fail to parse. It could be injected.
+		if err := checkResponseRaw(0, name, qtype, buf); err != nil {
```
fortuna, line 553: I think we should match on ID first, then validate.

fortuna (commented):
Some thoughts.

PR #596, opened 2026-04-18
feat(x): upgrade psiphon dependency

fortuna:


fortuna:
@greptile review

fortuna:
Ugh, this is a major pain. The issues in the Psiphon code is forcing me to litter our code with extra tags, and the build is still failing on some dependency error:

```
link: [github.com/wlynxg/anet](https://github.com/wlynxg/anet): invalid reference to net.zoneCache
```

It seems they are trying to link against an unexported variable that changed across Go versions.

I'm inclined to give up fixing Psiphon until their code is fixed.

/cc @ohnorobo

fortuna:
Found the issue and the potential fix: https://github.com/wlynxg/anet#how-to-build-with-go-1230-or-later

fortuna:
Nevermind, the fix is bad. It requires digging into the standard library internals, which is not guaranteed to be stable: https://tip.golang.org/doc/go1.23#linker

PR #597, opened 2026-04-27
Add README for lwip2transport

fortuna:
This should help us all better reason about the data flow in the Outline client. It's really hard to make sense of it.

It's been hard to work on the DNS leak issue and having to re-figure it out every time.

Also: there seems to be an opportunity to remove the app look in the inbound flow in favor of a callback that writes directly to the tun device. Something to consider later.


network/lwip2transport/README.md
```diff
@@ -0,0 +1,129 @@
...
+* The LwIP stack encapsulates the raw UDP payload into a full IP packet. The destination IP is determined by the local IP state held in the `UDPConn` association.
+
+#### 5. LwIP Outputs IP Packet
+Once LwIP constructs the IP packet, it calls its registered output callback: `forwardOutgoingIPPacket` (registered with `lwip.RegisterOutputFn`).
+
+#### 6. Sent to `lwIPDevice` Read Channel
+`forwardOutgoingIPPacket` writes the resulting IP packet into the `lwIPDevice`'s `d.rdBuf` channel. 
+
+#### 7. Read by App & Written to TUN
+The blocked `lwIPDevice.Read()` call (from Step 1) unblocks, copies the packet from `d.rdBuf`, and returns it to the application loop. The application then writes this complete IP packet back to the host OS local TUN interface.
```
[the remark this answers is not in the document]
fortuna: Fixed

network/lwip2transport/README.md
```diff
@@ -0,0 +1,129 @@
+# LwIP Packet Traversal Sequence
+
+This document tracks the traversal of packets through the `lwip2transport` LwIP network stack implementation, bridging the local host's TUN device and a remote proxy socket.
+
+## Architecture & State Management
+
+Before looking at the packet flows, it is important to understand how associations (connections) are managed between the TUN interface and the proxy.
+
```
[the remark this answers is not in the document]
fortuna: `udpConns` is correct: https://github.com/eycorsican/go-tun2socks/blob/301549c435ff1ca8fc353bee3763ae639c727b89/core/udp_callback_export.go#L33

fortuna:
FYI, the android issue has nothing to do with this PR. The emulator doesn't seem to have access to 8.8.8.8.

network/lwip2transport/README.md
```diff
@@ -0,0 +1,129 @@
...
+        Note over Handler: If new session:<br/>Calls proxy.NewSession(respWriter)<br/>Starts Proxy read loop.
+        Handler->>Proxy: WriteTo(payload, destAddr)
+        Proxy->>Remote: Send payload via Proxy socket
+    end
+```
+
+### Step-by-Step Walkthrough
+
+#### 1. Read from Local TUN
+The VPN application runs a continuous loop reading outbound IP packets from the host OS TUN device. (On platforms like Apple, the OS may call a callback instead of the app actively reading).
```
ohnorobo: nit: is the "may" here because there are multiple behaviors, or we're not sure? Maybe footnote an example if it's complicated.

ohnorobo (approved):

network/lwip2transport/README.md
```diff
@@ -0,0 +1,129 @@
+# LwIP Packet Traversal Sequence
+
+This document tracks the traversal of packets through the `lwip2transport` LwIP network stack implementation, bridging the local host's TUN device and a remote proxy socket.
```
jyyi1, line 3: *nit* TUN device is just one of the supported devices, it basically supports any `io.ReadWriter`. But it's OK to use TUN in this example.

network/lwip2transport/README.md
fortuna: Updated it to clarify.

network/lwip2transport/README.md
fortuna: I think TUN makes it clearer to the user here.

PR #625, opened 2026-06-16
fix(packetrelay): add PacketListenerRelay idle deadlines

fortuna:
## Summary

- Give every `PacketListenerRelay` association a default 30s write-idle timeout using `net.PacketConn` deadlines.
- Remove `NewPacketRelayFromPacketListener` constructor options; customize the timeout with `PacketListenerRelay.SetWriteIdleTimeout`.
- Update the deprecated `PacketListenerProxy` adapter to use the native `PacketListenerRelay` timeout instead of wrapping with `TimeoutPacketRelay`.
- Treat IPv4 and IPv4-mapped IPv6 addresses as equivalent when matching the DNS intercept local resolver.

## Why

`PacketListenerRelay` associations need a timeout wherever they are used; otherwise dropped UDP packets can leave associations and goroutines alive indefinitely. Call sites were having to remember to wrap every `PacketListenerRelay` in `TimeoutPacketRelay`, which made this easy to miss.

The DNS intercept issue exposed another compatibility gap: some network stacks surface IPv4 UDP destinations in IPv4-mapped IPv6 form. Exact `netip.AddrPort` equality meant a configured resolver like `10.0.0.1:53` did not match `[::ffff:10.0.0.1]:53`, so DNS queries could fall through to the default relay and time out.

## Testing

- `go test ./network/packetrelay`
- `go test ./network`
- `go test ./network/dnsintercept`
- `go test ./...`

### DIFF
--- a/network/packetrelay/packet_listener_relay.go
+++ b/network/packetrelay/packet_listener_relay.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"net/netip"
 	"sync"
+	"time"
 
 	"golang.getoutline.org/sdk/internal/slicepool"
 	"golang.getoutline.org/sdk/transport"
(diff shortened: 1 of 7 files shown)


network/packetrelay/packet_listener_relay.go
fortuna, line 1: TODO: Make SetRelay in the delegate relay not return an error. It's another API update we need to make for the client.

jyyi1 (approved):
LGTM, thanks!

PR #627, opened 2026-06-16
test(x): mock StrategyFinder dialers in smart integration tests

fortuna:
Simplify smart integration tests by using mock dialers and an error fallback parser. This removes the need for the nettest build tag and external network connections.


ohnorobo (approved):
thank you for the flakiness fix!

PR #634, opened 2026-09-08
feat(fetch): add QUIC version selection

fortuna:
### Summary

- add an ordered `-quic-versions` option to the fetch CLI for HTTP/3
- support numeric selection of QUIC versions 1 and 2
- report the negotiated QUIC version and wire codepoint in verbose output
- add parser coverage and document the new option

### Motivation

The fetch tool currently cannot directly test QUIC v2 reachability or compare
version-specific network behavior. Applying the requested order to the HTTP/3
connection configuration allows forced-version and version-negotiation tests
while preserving the existing default preference.

### Testing

- `go test ./tools/fetch` from the `x` module
- forced v1 and v2 HTTP/3 requests against Caddy v2.11.4
- forced v1 and v2 requests against QUIC-Go's public interoperability server

### DIFF
--- a/x/tools/fetch/README.md
+++ b/x/tools/fetch/README.md
@@ -2,16 +2,16 @@
 
 This app illustrates how to use different transports to fetch a URL in Go.
 
-When using HTTP/3, `-quic-versions` selects the ordered QUIC versions offered by
-QUIC-Go. For example, force RFC 9369 QUIC v2 with:
+When using HTTP/3, `-quic-versions` selects the ordered QUIC versions. For
+example, force RFC 9369 QUIC v2 with:
 
 ```console
-fetch -proto h3 -quic-versions v2 -method HEAD -v https://example.com/
+fetch -proto h3 -quic-versions 2 -method HEAD -v https://example.com/
 ```
 
 The tool logs the negotiated QUIC version and wire codepoint after a successful
-HTTP/3 request. Use `v1,v2` (the default) for QUIC-Go's normal preference order,
-or `v2,v1` to prefer v2 while allowing fallback through version negotiation.
+HTTP/3 request. Use `1,2` (the default) to prefer v1, or `2,1` to prefer v2
+while allowing fallback through version negotiation.
 
 Direct fetch:
 
--- a/x/tools/fetch/README.md
+++ b/x/tools/fetch/README.md
@@ -2,6 +2,17 @@
 
 This app illustrates how to use different transports to fetch a URL in Go.
 
+When using HTTP/3, `-quic-versions` selects the ordered QUIC versions offered by
+QUIC-Go. For example, force RFC 9369 QUIC v2 with:
+
+```console
+fetch -proto h3 -quic-versions v2 -method HEAD -v https://example.com/
+```
+
+The tool logs the negotiated QUIC version and wire codepoint after a successful
+HTTP/3 request. Use `v1,v2` (the default) for QUIC-Go's normal preference order,
+or `v2,v1` to prefer v2 while allowing fallback through version negotiation.
+
 Direct fetch:
 
 ```sh
--- a/x/tools/fetch/main.go
+++ b/x/tools/fetch/main.go
@@ -40,7 +40,7 @@ import (
 
 type stringArrayFlagValue []string
 
-const defaultQUICVersions = "v1,v2"
+const defaultQUICVersions = "1,2"
 
 func (v *stringArrayFlagValue) String() string {
 	return fmt.Sprint(*v)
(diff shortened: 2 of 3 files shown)


x/tools/fetch/README.md
```diff
@@ -2,6 +2,17 @@
 
 This app illustrates how to use different transports to fetch a URL in Go.
 
+When using HTTP/3, `-quic-versions` selects the ordered QUIC versions offered by
+QUIC-Go. For example, force RFC 9369 QUIC v2 with:
```
fortuna: We shouldn’t leak the QUIC-go implementation here

x/tools/fetch/README.md
```diff
@@ -2,6 +2,17 @@
...
+
+```console
+fetch -proto h3 -quic-versions v2 -method HEAD -v https://example.com/
+```
+
+The tool logs the negotiated QUIC version and wire codepoint after a successful
+HTTP/3 request. Use `v1,v2` (the default) for QUIC-Go's normal preference order,
```
fortuna: Let’s use just the numbers for the version selection (1,2, not v1,v2)

x/tools/fetch/README.md
fortuna: Updated the README to describe the option without naming the underlying QUIC implementation.

x/tools/fetch/README.md
fortuna: Updated the flag, parser, tests, and README to use numeric selectors only (`1`, `2`, `1,2`, and `2,1`).

x/tools/fetch/main.go
```diff
@@ -113,6 +138,10 @@ func main() {
 		flag.Usage()
 		os.Exit(1)
 	}
+	if *protoFlag != "h3" && *quicVersionsFlag != defaultQUICVersions {
+		slog.Error("-quic-versions requires -proto h3")
+		os.Exit(1)
+	}
```
angelodlfrtr: `go run main.go -v -method HEAD -proto h1 -quic-versions "1,2" ...` does not return an error here, because `"1,2" == defaultQUICVersions`. But not a blocking issue.

x/tools/fetch/main.go
```diff
@@ -113,6 +138,10 @@ func main() {
 		flag.Usage()
 		os.Exit(1)
 	}
+	if *protoFlag != "h3" && *quicVersionsFlag != defaultQUICVersions {
```
angelodlfrtr:
```suggestion
	if *protoFlag != "h3" && *quicVersionsFlag != "" {
```

x/tools/fetch/main.go
```diff
@@ -80,6 +104,7 @@ func main() {
 	transportFlag := flag.String("transport", "", "Transport config")
 
 	protoFlag := flag.String("proto", "h1", "HTTP version to use (h1, h2, h3)")
+	quicVersionsFlag := flag.String("quic-versions", defaultQUICVersions, "Ordered QUIC versions for h3 (1, 2, or a comma-separated list)")
```
angelodlfrtr:
```suggestion
	quicVersionsFlag := flag.String("quic-versions", "", fmt.Sprintf("Ordered QUIC versions for h3 (1, 2, or a comma-separated list) (default %s)", defaultQUICVersions))
```

x/tools/fetch/main.go
```diff
@@ -174,6 +204,11 @@ func main() {
 			}
 		}
 	} else if *protoFlag == "h3" {
+		quicVersions, err := parseQUICVersions(*quicVersionsFlag)
```
angelodlfrtr:
```suggestion
		quicVersionValue := *quicVersionsFlag
		if quicVersionValue == "" {
			quicVersionValue = defaultQUICVersions
		}
		quicVersions, err := parseQUICVersions(quicVersionValue)
```

angelodlfrtr (approved):

x/tools/fetch/main.go
fortuna: Fixed in 63847ee. The flag now uses an empty sentinel, so explicitly passing `-quic-versions 1,2` with h1/h2 is rejected. Verified with the built CLI and `go test ./tools/fetch`.

x/tools/fetch/main.go
fortuna: Applied in 63847ee: protocol validation now checks for a non-empty `-quic-versions` value.

x/tools/fetch/main.go
fortuna: Applied in 63847ee: the flag default is empty while the help text continues to show the effective `1,2` default.

x/tools/fetch/main.go
fortuna: Applied in 63847ee: HTTP/3 maps an omitted flag to `defaultQUICVersions` before parsing.

PR #638, opened 2026-09-16
feat(quicprelude): QUIC preludes as a package and transport config

fortuna:
## Summary

- add `x/quicprelude`, a package that builds prelude datagrams and produces a packet listener that sends them ahead of every QUIC ClientHello, with an injectable generator
- register it in `configurl` as `quicprelude`, so it composes with the rest of the transport stack
- have the `fetch` tool use it, configured through `-transport`

## Motivation

Some middleboxes read the TLS SNI from the first QUIC Initial they can parse on a four-tuple, cache a verdict for that flow, and apply it to everything after. An Initial-shaped datagram they cannot decrypt yields no name, so no verdict is cached and the real request is not matched against one.

Building this into `fetch` as flags would make it measurement-only: no other SDK consumer could use it, and `fetch` would own both packet construction and dial-time sequencing. As a transport it is available to anything that builds a packet listener from a config, and the prelude shares a four-tuple with the traffic it precedes by construction rather than by careful wiring.

## API

The zero value of a prelude listener is not valid — it has no inner listener and
no datagram source — so by the repository's construction patterns the
implementation stays unexported and is reached through a config factory that
returns the interface:

```go
config := quicprelude.NewConfig()
listener, err := config.NewPacketListener(inner)
```

`NewConfig` is infallible; errors surface in `NewPacketListener`, where the
dependency is needed. The factory copies the settings, so configuring a `Config`
after creating a listener does not reach that listener.

Datagrams come from a `Generator`, which is handed the packet it is about to
precede:

```go
type GeneratorInput struct {
    Packet      []byte
    Destination net.Addr
}

type Generator func(input GeneratorInput) ([][]byte, error)
```

The arguments are a struct so fields can be added later without breaking
implementations.

Seeing the packet lets a generator match its length, so the prelude is not
separable by size from the traffic it is mixed with; `MatchPacketLength` is the
default. It also makes the ClientHello available, so a generator can read the
SNI out of an Initial and vary on it.

Returning a list covers shapes a single datagram could not: a split Initial, a
varying burst, or none at all. Returning none sends the packet unchanged.

`InvalidInitial` and `Random` are ordinary generator constructors and `Repeat`
is a combinator, so the set of shapes is not closed by an enum — a caller who
needs something else writes a function.

```go
version, err := quicprelude.FixedVersion(quicprelude.Version2)
generator, err := quicprelude.InvalidInitial(version, quicprelude.MatchPacketLength)
generator, err = quicprelude.Repeat(2, generator)
listener, err := quicprelude.NewConfig().WithGenerator(generator).NewPacketListener(inner)
```

Both constructors validate eagerly rather than deferring to the first write,
which is what lets `configurl` reject a bad length at parse time.

## When a prelude is sent

Before every datagram that may carry a QUIC ClientHello, decided from the packet
alone. The listener keeps no per-destination state.

An earlier revision sent one prelude per destination for the life of the
socket. That preluded whatever reached a new address first, DNS included, and
left a second QUIC connection on a reused socket without one, even after a
middlebox had expired its flow state. It also needed a bounded map and a lock
held across writes.

Only an Initial carrying a ClientHello holds a name a filter can read. Knowing
for certain would mean decrypting it, so the listener reads the header: a client
Initial of v1, a draft version or v2 whose header parses, in a datagram of at
least 1200 bytes (the minimum RFC 9000 allows a client Initial), gets a prelude, unless a Handshake
packet is coalesced after it, which is how some stacks send the ACK of the server's
Initial. Anything that is not such an Initial, including all non-QUIC traffic,
passes untouched. Retransmissions, both halves of a post-quantum ClientHello
split across two packets, and the resend after a Retry, a HelloRetryRequest or
Version Negotiation each get their own prelude. The rule errs toward an extra
prelude, never toward a ClientHello without one.

Measured against `www.google.com` with QUIC-Go v0.48.1 on Go 1.25: the default
key share splits the ClientHello across two Initials, and QUIC-Go sends the ACK
as a standalone padded Initial rather than coalescing it, so a handshake sends
three preludes, one of them unnecessary. With X25519 only, it sends two.

## Configuration

```
quicprelude:count=[COUNT]&mode=[MODE]&length=[LENGTH]&version=[VERSION]
```

Every option may be omitted, and a bare `quicprelude` is the configuration to
reach for first.

| Option | Default | Meaning |
|---|---|---|
| `count` | `1` | Number of datagrams, at most 16; `0` disables the prelude |
| `mode` | `invalid-initial` | Or `random`, useful as a control |
| `length` | `match` | Sizes each datagram like the packet it precedes; or a byte count |
| `version` | `0x1a2a3a4a` | A reserved codepoint; also `v1`, `v2`, or any hex value |

It stacks above a proxy, so the prelude travels the same path as what follows:

```console
fetch -proto h3 -method HEAD -v \
  -transport 'socks5://user:pass@proxy.example:1080|quicprelude:' \
  https://example.com/
```

## The defaults work unmodified in both regimes measured

Russia filters on the SNI in the first parseable Initial; Iran filters on the
version codepoint. A bare `quicprelude`, with no options, covers both. Measured at `46103f3`, which preludes every
ClientHello, forced v1 to `www.youtube.com`, arms shuffled every round:

| Vantage points | Control | Bare `quicprelude` |
|---|---|---|
| Russia, five sessions on four networks (Rostelecom, MTS, TransTeleCom, MegaFon ×2), pinned destination, ten rounds each | 3 / 50 | 49 / 50 |
| Iran, six exits on three networks (MCI, Aria Shatel, AS34369), five rounds each | 0 / 30 | 27 / 30 |

Against the previous one-prelude-per-destination build (`929a18b`), in an
interleaved follow-up on MegaFon and TransTeleCom, preluding every ClientHello
did no worse: 61/80 against 59/80.

One result we cannot explain: in Russia, a ClientHello that fits in a single
Initial is cleared less reliably than one split across two, 55/80 against 65/80
across both builds (`GODEBUG=tlsmlkem=0` produces the single-packet case). The
split case includes a prelude between the two halves, so that is not what hurts.
It matters for clients without a post-quantum key share.

## Why a reserved version is the default

Measurements on Russian and Iranian vantage points found that filtering which drops recognized QUIC versions passes codepoints it does not know, and that v2 is already dropped on some paths. A codepoint no implementation speaks has no reason to appear in such a list. QUIC v2 with correct Initial type bits also works today; the reserved codepoint is the more durable default.

Packet structure matters as much as the number: in Russia, random bytes of the same length had no effect, and the same v2 codepoint worked with Initial type bits and failed with Retry type bits.

## Testing

- `go test -C x ./quicprelude ./configurl ./tools/fetch`
- `CGO_ENABLED=1 go test -C x -race ./quicprelude ./configurl ./tools/fetch`
- `go -C x tool staticcheck ./quicprelude ./configurl ./tools/fetch`
- a live QUIC-Go handshake to `www.google.com` with every client datagram logged, confirming that only client Initials are preluded (three per handshake, as above) and that HTTP/3 still completes over v1 and over `2,1`
- a loopback UDP sink confirming the preludes precede the real Initial on the same source port, with the expected version and type bits
- on a Russian vantage point, one pinned destination, control and treatment interleaved in the same session: control 0/8, prelude 8/8; re-verified after each API rework, most recently control 0/6, prelude 6/6

## Notes for review

- The ClientHello test reads the header rather than decrypting the Initial. Decrypting would skip the unnecessary prelude before QUIC-Go's standalone ACK, but it means reimplementing the Initial key schedule for each version. Say if you think it is worth it.
- Preluding every ClientHello has been validated against both filters (above). Preludes ahead of a retransmitted ClientHello, or the resend after Retry or HelloRetryRequest, were not isolated.
- This is the first packet **listener** wrapper in `configurl`; `override` wraps dialers only.
- An earlier commit on this branch (`93dca54`) committed a 12 MB binary (`x/fetch`) by accident. It is removed, but it remains in the branch history; say the word and I will rewrite the branch to drop the blob.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_019dNmgsvgSiLsFKkianPuyr


.gitignore
```diff
@@ -26,4 +26,12 @@ output-metadata.json
 
 # Tasks
 .task
-out
\ No newline at end of file
+out
+
+# Binaries built in place by `go build ./tools/<name>`, which lands the
```
fortuna: Revert this. You can use go run, no need to build

fortuna (commented):
Remove the reference to valid-v2 in the pr description. That doesn't exist.

.gitignore
fortuna: Reverted in 785f439. The binary itself stays removed; there is no in-tree build to ignore.

fortuna:
Description rewritten: the valid-v2 section and the summary bullet are gone. You are right that it never existed on main, so describing its removal only raised a question the diff cannot answer.

Also reverted the .gitignore entries (785f439).

fortuna:
Done in 4284b42. Replaced the tables with explicit calls and assertions, and switched to testify `require` to match the rest of the module.

You were right that it was foreign here rather than just harder to read: testify appears in 24 test files across `x/`, and the only files using a table were the ones this branch added. Failure output now names the case instead of an index.

I converted `listener_test.go` the same way, so the tests added here are consistent with each other and with their neighbours. Same coverage: 27 tests, `-race` and `staticcheck` clean.

fortuna:
@greptile review

x/quicprelude/quicprelude.go
```diff
@@ -0,0 +1,342 @@
...
+
+// Random returns a [Generator] producing opaque random bytes. A middlebox that
+// parses QUIC will not recognize them as QUIC at all, which makes this useful
+// as a control rather than as a technique.
+func Random(length int) (Generator, error) {
+	if length < 0 {
+		return nil, fmt.Errorf("length must not be negative, got %d", length)
+	}
+	return func(input GeneratorInput) ([][]byte, error) {
+		datagram, err := randomBytes(lengthFor(length, input.Packet, 1))
```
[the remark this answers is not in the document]
fortuna: Fixed in 46103f3, though not with the suggested fallback. `Random` is the size-matched control, so with `length=match` it now matches the packet exactly, whatever its length: random bytes have no Initial-sized minimum, and substituting `DefaultLength` for a short packet would still make the control stand out. Only an empty packet, which has no length to match, falls back to `DefaultLength`. Tests cover 40 bytes and empty. (Since add9013 the listener only preludes client Initials, so short packets reach a generator only when an Initial is malformed, but the generator should be right on its own.)

x/quicprelude/quicprelude.go
```diff
@@ -0,0 +1,342 @@
...
+//
+//	config := quicprelude.NewConfig()
+//	listener, err := config.NewPacketListener(inner)
+//
+// The datagrams come from a [Generator], which is handed the packet it is about
+// to precede. [InvalidInitial] and [Random] cover the cases this package was
+// written for, [Repeat] sends one of them several times, and a caller who needs
+// something else supplies their own:
+//
+//	generator, err := quicprelude.InvalidInitial(quicprelude.Version2, quicprelude.MatchPacketLength)
```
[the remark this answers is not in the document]
fortuna: Right, the example predates `VersionSource`. Fixed in 46103f3 by wrapping the version with `FixedVersion`. Both documented examples are now `Example` functions in `example_test.go`, so they compile with the tests and cannot drift again.

x/configurl/quicprelude.go
```diff
@@ -0,0 +1,150 @@
...
+	for key, vs := range values {
+		if len(vs) != 1 {
+			return nil, fmt.Errorf("option %v must have exactly one value, found %v", key, len(vs))
+		}
+		value := vs[0]
+		switch strings.ToLower(key) {
+		case "count":
+			if options.count, err = strconv.Atoi(value); err != nil {
+				return nil, fmt.Errorf("invalid count %q: %w", value, err)
+			}
```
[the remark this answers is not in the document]
fortuna: Agreed. Fixed in 46103f3: `Repeat` rejects counts above `MaxRepeat` (16), and `configurl` inherits the bound, with tests for 16 and 17. I bounded it rather than generating incrementally because the datagrams are sent under a lock ahead of the packet, and no measurement needed more than three.

x/quicprelude/config.go
```diff
@@ -0,0 +1,152 @@
...
+
+// maxTrackedDestinations bounds the per-connection record of destinations that
+// have already been preluded. A long-lived connection that exceeds it starts
+// over, which resends a prelude to a destination already seen. That is wasteful
+// but harmless, and it keeps memory bounded.
+const maxTrackedDestinations = 1024
+
+// Config describes the datagrams to send ahead of a flow's real traffic, and
+// produces the listener that sends them. Its zero value sends nothing.
+//
```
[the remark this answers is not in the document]
fortuna: The comment was wrong. Fixed in 46103f3: it now says to create a `Config` with `NewConfig`, and that `NewPacketListener` rejects a zero `Config`. A silent no-op zero value would hide a misconfiguration, so I kept the error.

fortuna:
Addressed all of the review in 46103f3, including the outside-diff one: `SanitizeConfig` now keeps `quicprelude` options, which carry no secrets, instead of logging `quicprelude://UNKNOWN`, with a test.

Note that the summary above describes per-destination tracking. That was replaced in add9013: the listener now preludes every datagram that may carry a ClientHello and keeps no state (see "When a prelude is sent" in the description).

@greptile review

x/quicprelude/quicprelude.go
```diff
@@ -0,0 +1,366 @@
...
+
+// RandomReservedVersion returns a [VersionSource] yielding a fresh codepoint
+// from the range RFC 9000, Section 15 reserves "for use in forcing version
+// negotiation to be exercised", 0x?a?a?a?a. The same section says a client "MAY
+// use one of these version numbers with the expectation that the server will
+// initiate version negotiation", so sending one is sanctioned behavior.
+//
+// This is the default. The range holds 65536 values, none of which can collide
+// with an assigned version, since it is reserved.
+func RandomReservedVersion() VersionSource {
```
angelodlfrtr, line 202: As of Go 1.24, crypto/rand.Read cannot fail.

x/quicprelude/quicprelude.go
```diff
@@ -0,0 +1,366 @@
...
+	if length < MinimumInitialLength {
+		return fmt.Errorf("length must be at least %d bytes, got %d", MinimumInitialLength, length)
+	}
+	if length-headerLength >= maxProtectedLength {
+		return fmt.Errorf("length must be under %d bytes, got %d", headerLength+maxProtectedLength, length)
+	}
+	return nil
+}
+
+func randomBytes(length int) ([]byte, error) {
```
angelodlfrtr, line 331: Idem, `rand.Read` cannot fail.

x/quicprelude/quicprelude.go
```diff
@@ -0,0 +1,366 @@
...
+// The floor matters. Filtering recognizes the versions that saw deployment: on
+// the Iranian paths measured for this package, draft-29 is dropped outright
+// while an unassigned codepoint sharing its prefix passes, so that filter
+// matches an exact list rather than the prefix. Choosing above the assigned
+// drafts stays clear of the list while remaining in a range a middlebox still
+// recognizes as QUIC.
+//
+// This is an alternative to [RandomReservedVersion] for a path where the
+// reserved range is filtered. It is a smaller pool.
+func RandomDraftVersion() VersionSource {
```
angelodlfrtr, line 228: Idem, `rand.Read` cannot fail.

x/quicprelude/quicprelude.go
```diff
@@ -0,0 +1,366 @@
...
+			return nil, err
+		}
+		return [][]byte{datagram}, nil
+	}, nil
+}
+
+// VersionSource chooses the version codepoint for a datagram. It is called once
+// per datagram, so a source that varies produces datagrams that differ on the
+// wire and gives a middlebox no single constant to match.
+type VersionSource func() (uint32, error)
```
angelodlfrtr, line 184: Since all implementations relies on `rand.Read` that cannot fail : 

```suggestion
type VersionSource func() uint32
```

Unless it was for a third party implementation that could fail ?

angelodlfrtr: https://go.dev/doc/go1.24#cryptorandpkgcryptorand

x/quicprelude/config.go
```diff
@@ -0,0 +1,131 @@
...
+// ClientHello. Everything else is delegated to the embedded [net.PacketConn].
+//
+// It keeps no record of destinations. Deciding from the packet alone means a new
+// connection on a reused socket gets a prelude like the first one did, including
+// one made after a middlebox has forgotten the flow, and that traffic which is
+// not QUIC never gets one. The cost is a prelude before each retransmitted or
+// resent ClientHello, which is also when a lost prelude most needs replacing,
+// and before some Initials that only acknowledge the server's. See
+// [mayCarryClientHello] for which datagrams qualify.
+type preludeConn struct {
```
angelodlfrtr, line 100: Embedding `net.PacketConn` as an interface is what makes this package work, and I
think that deserves to be written down here, because the obvious "improvement"
silently breaks it.

Only the seven methods `net.PacketConn` declares get promoted, so `SyscallConn`,
`SetReadBuffer`, `ReadMsgUDP` and `WriteMsgUDP` are not reachable on `*preludeConn`
even when a real `*net.UDPConn` is underneath. QUIC-Go probes for exactly those in
`wrapConn` (`sys_conn.go:29-76`, v0.48.1) and falls back to `basicConn` when any is
missing — and `basicConn.WritePacket` sends with `PacketConn.WriteTo`. That
fallback is the only reason `WriteTo` ever sees a ClientHello. The OOB path sends
with `WriteMsgUDP`, which would reach the inner conn directly: no prelude, no
error, handshake still succeeds, and only a packet capture would show it.

So forwarding those four methods — a five-line change that reads as a pure
optimization, and that someone will eventually propose after seeing the
receive-buffer warning — disables the technique with every test still green.

<details>
<summary>What the wrapping currently costs, per platform</summary>

| GOOS | Cost of wrapping | Would forwarding break the prelude? |
|---|---|---|
| linux (→ android) | GSO + ECN + DF/PMTU | **yes** |
| darwin (→ ios) | ECN + DF/PMTU (no GSO on darwin) | **yes** |
| freebsd | ECN only | **yes** |
| windows | DF/PMTU only | no — `sys_conn_windows.go` returns `basicConn` regardless |
| others | nothing | no |

`isGSOEnabled` is real only in `sys_conn_helper_linux.go`; darwin and freebsd
hardcode `false`. Windows is immune in v0.48.1 only because it has no OOB send
path yet — worth not relying on across a bump.

One knock-on: since DF is never set, `EMSGSIZE` (the one send error QUIC-Go
tolerates rather than treating as fatal) can't arise on a preluded conn. That makes
the fail-the-write behaviour in `WriteTo` below safer than it first looks.

</details>

Two suggestions:

1. A paragraph on this doc comment recording the trade and warning against
   forwarding.
2. A test that fails if anyone does. I wrote one — it wraps a fake conn that *does*
   carry the four methods, asserts the wrapper hides them, and separately asserts
   `WriteMsgUDP` is unreachable (the composite check alone goes stale if QUIC-Go
   ever narrows its probe). Verified it fails when the forwarding methods are added,
   and it cross-compiles on linux/darwin/windows/freebsd/wasm/plan9, with android
   and ios type-checking clean.

Assisted-by: Claude Code (Claude Opus 5)

x/quicprelude/quicprelude.go
```diff
@@ -0,0 +1,366 @@
...
+//	generator, err := quicprelude.InvalidInitial(version, quicprelude.MatchPacketLength)
+//	generator, err = quicprelude.Repeat(2, generator)
+//	listener, err := quicprelude.NewConfig().
+//		WithGenerator(generator).
+//		NewPacketListener(inner)
+//
+// The datagrams [InvalidInitial] produces are deliberately not valid QUIC. They
+// carry a long header with a plausible version and connection IDs, and random
+// bytes where the protected payload and authentication tag would be. A QUIC
+// server discards them.
```
angelodlfrtr: I think this is inverted, and it matters for how someone reasons about the
technique on a path you haven't measured.

A server never reaches the AEAD step. It reads the version field first, and
RFC 9000 §6 requires it to answer a version it doesn't support with a Version
Negotiation packet. §15 says the `0x?a?a?a?a` range exists specifically to force
that behaviour — which is exactly what makes it defensible to send one, and also
what obliges a compliant server to reply.

QUIC-Go does it at `server.go:400-416`, gated on `MinUnknownVersionPacketSize`,
which is `MinInitialPacketSize` = 1200. The prelude is always ≥1200 by
construction, so it clears the threshold every time — the size floor that makes
`mayCarryClientHello` safe is the same floor that guarantees the reply. A default
handshake sends three preludes, so three VN packets come back, each echoing the
decoy's connection IDs with the roles swapped
(`internal/wire/version_negotiation.go:35-47`).

None of this contradicts your measurements: the filters you tested clearly aren't
watching the return path. But "no server acts on it" is what someone will reason
from when deciding whether this is safe somewhere you haven't been, and the return
path is where a second, independent signal exists. I'd rather the doc said what
actually happens and left the judgement to the reader.

Assisted-by: Claude Code (Claude Opus 5)

x/quicprelude/config.go
```diff
@@ -0,0 +1,131 @@
...
+// preludeConn sends the prelude before every datagram that may carry a QUIC
+// ClientHello. Everything else is delegated to the embedded [net.PacketConn].
+//
+// It keeps no record of destinations. Deciding from the packet alone means a new
+// connection on a reused socket gets a prelude like the first one did, including
+// one made after a middlebox has forgotten the flow, and that traffic which is
+// not QUIC never gets one. The cost is a prelude before each retransmitted or
+// resent ClientHello, which is also when a lost prelude most needs replacing,
+// and before some Initials that only acknowledge the server's. See
+// [mayCarryClientHello] for which datagrams qualify.
```
angelodlfrtr, line 99:
```suggestion
// [mayCarryClientHello] for which datagrams qualify.
//
// The [net.PacketConn] is embedded as an interface, and deliberately so. Only
// the seven methods that interface declares are promoted; the ones a
// [net.UDPConn] additionally has are not, whatever the connection underneath
// turns out to be. A QUIC stack probes for those to find the faster paths.
// QUIC-Go asks for SyscallConn, SetReadBuffer, ReadMsgUDP and WriteMsgUDP, and
// when any is missing falls back to a connection that sends with WriteTo. That
// fallback is what lets this wrapper see a ClientHello at all: the faster path
// sends with WriteMsgUDP, which would reach the inner connection directly and
// leave the prelude unsent, with no error to report it and no test to catch it.
//
// The price is that a preluded connection runs without segmentation offload,
// without ECN, and without the DF bit that path MTU discovery needs, so its
// packets never grow past the stack's conservative initial size. That is the
// cost of the technique rather than an oversight, and forwarding the inner
// connection's methods to recover it would silently disable the prelude. The
// test TestPreludeConnHidesOOBMethods fails if they are ever forwarded.
```

Assisted-by: Claude Code (Claude Opus 5)

x/quicprelude/config_test.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+type fixedListener struct {
+	conn net.PacketConn
+}
+
+func (l *fixedListener) ListenPacket(context.Context) (net.PacketConn, error) {
+	return l.conn, nil
+}
+
```
angelodlfrtr, line 90:
```go
import "syscall"

// oobCapablePacketConn is the shape QUIC-Go probes for before it chooses the
// send path that bypasses WriteTo. It is declared here rather than imported so
// the package keeps no dependency on a QUIC implementation: what matters is
// that the prelude connection does not have these methods, not which interface
// they happen to belong to.
type oobCapablePacketConn interface {
	net.PacketConn
	SyscallConn() (syscall.RawConn, error)
	SetReadBuffer(int) error
	ReadMsgUDP(b, oob []byte) (n, oobn, flags int, addr *net.UDPAddr, err error)
	WriteMsgUDP(b, oob []byte, addr *net.UDPAddr) (n, oobn int, err error)
}

// oobCapableConn records writes like a [recordingConn] and additionally carries
// the methods of a [net.UDPConn] that a QUIC stack looks for, so a test can wrap
// a connection that really does offer the faster paths.
type oobCapableConn struct {
	recordingConn
}

func (*oobCapableConn) SyscallConn() (syscall.RawConn, error) {
	return nil, errors.New("not implemented")
}

func (*oobCapableConn) SetReadBuffer(int) error { return nil }

func (*oobCapableConn) ReadMsgUDP([]byte, []byte) (int, int, int, *net.UDPAddr, error) {
	return 0, 0, 0, nil, errors.New("not implemented")
}

func (*oobCapableConn) WriteMsgUDP([]byte, []byte, *net.UDPAddr) (int, int, error) {
	return 0, 0, errors.New("not implemented")
}

// TestPreludeConnHidesOOBMethods guards the invariant the technique rests on. A
// QUIC stack that finds SyscallConn, SetReadBuffer, ReadMsgUDP and WriteMsgUDP
// on its connection sends with WriteMsgUDP, which would reach the inner
// connection without passing through WriteTo and leave every prelude unsent.
// Nothing would fail: the handshake would still complete, and only a packet
// capture would show the prelude missing. So forwarding those methods to the
// connection underneath, however reasonable it looks as an optimization, must
// not compile past this test. See the note on preludeConn.
func TestPreludeConnHidesOOBMethods(t *testing.T) {
	inner := &oobCapableConn{}
	// The inner connection offers the faster paths, so this test is not vacuous.
	require.Implements(t, (*oobCapablePacketConn)(nil), inner)

	listener, err := NewConfig().NewPacketListener(&fixedListener{conn: inner})
	require.NoError(t, err)
	conn, err := listener.ListenPacket(context.Background())
	require.NoError(t, err)

	require.NotImplements(t, (*oobCapablePacketConn)(nil), conn,
		"the prelude connection must not offer a send path that bypasses WriteTo")

	// Asserted separately, because the composite check above only holds while a
	// QUIC stack asks for all four methods. WriteMsgUDP is the one that actually
	// sends, so a connection carrying it can be written through whatever shape a
	// future version of a stack decides to probe for.
	type msgWriter interface {
		WriteMsgUDP(b, oob []byte, addr *net.UDPAddr) (n, oobn int, err error)
	}
	require.NotImplements(t, (*msgWriter)(nil), conn,
		"WriteMsgUDP must not be reachable on the prelude connection")

	// And the prelude still reaches the wire through WriteTo.
	_, err = conn.WriteTo(clientInitial(1200), udpAddr(t, "192.0.2.1:443"))
	require.NoError(t, err)
	require.Equal(t, 1, inner.countPreludes())
}
```

Assisted-by: Claude Code (Claude Opus 5)

angelodlfrtr (commented):
I think this PR is mergeable as this.

The claude findings are for reference, nothing that block.

Just ensure removing properly the fetch binary from the history before merge.

